PR #5, opened 2022-06-10
Fix rawhide builds due to wrong chroot name

nikic:
The rawhide chroots are called fedora-rawhide-x86_64 or similar,
not fedora-37-x86_64. As we already know the exact chroot name,
just use it directly rather than rederiving it.


kwk (approved):
LGTM

PR #8, opened 2022-09-06
Test new method of cleaning and forking in copr

kwk:
Rework how we build snapshots

When we build a snapshot we want to build it with the clang that
ships with Fedora. What happened before was that we've build clang with
the latest version of clang that we've built ourselves. This is of
course unfortunate.

This is an approach to have three copr projects:

0. `llvm-snapshots-incubator-<YESTERDAY>` (will be kept for 7 days)
1. `llvm-snapshots-incubator-<TODAY>` (will be kept for 7 days)
2. `llvm-snapshots` (will be overwritten each day with `llvm-snapshots-incubator-<TODAY>`)

The `llvm-snapshots` project is *forked* from the
`llvm-snapshot-incubator-<YESTERDAY>` over and over again every day. This copies
over successful builds from the `llvm-snapshots-incubator-<YESTERDAY>` project.

The advantages are:

 * Every day we build in a fresh project without any packages from previous builds.
 * We can look back 7 days and see projects.
 * We don't have to tell users to change how they get the snapshots.

Downsides:

 * Collecting statistics need to be adjusted.
 * Lot more juggling with projects.

I hope this makes sense and looks right.

NOTICE: For now I'll do this under my ownername (`kkleine` and not under `@fedora-llvm-team`)

### DIFF
--- a/.github/workflows/fedora-copr-build.yml
+++ b/.github/workflows/fedora-copr-build.yml
@@ -31,14 +31,18 @@ jobs:
         with:
           python-version: 3.9
 
-      - uses: actions/checkout@v2
+      - name: "checkout this repo"
+        uses: actions/checkout@v2
 
-      - run: |
+      - name: "install python requirements"
+        run: |
           python -m pip install --upgrade pip
           pip install -r requirements.txt
-        
-      - name: copr build
-        env:
+
+      - name: "prepare variables"
+        # We specify the secrets here and give them an alias for other steps to
+        # repeat the options.
+        env: &secrets
           # You need to have those secrets in your repo.
           # See also: https://copr.fedorainfracloud.org/api/.
           COPR_LOGIN: ${{ secrets.COPR_LOGIN }}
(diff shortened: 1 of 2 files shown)


Nanotwerp:
> When we build a snapshot of clan we want to build it with the clang that
> ships with Fedora. What happened before was that we've build clang with
> the latest version of clang that we've built ourselves. This is of
> course unfortunate.

Is this why builds of Clang have been failing for a while? I thought something terribly wrong happened, haha!

praiskup:
> @praiskup what do you think? Does this reflect what we've been writing about today?

Yes, that sounds good.

> The problems I see is that once we kick off the fork, I think we have to
wait until that is finished.

IIUC, your build batches build longer than 24 hours ... which in turn means that you can't really build _daily_ anymore?

praiskup:
Unless you have more than one "incubator" project.

kwk:
> > When we build a snapshot of clan we want to build it with the clang that
> > ships with Fedora. What happened before was that we've build clang with
> > the latest version of clang that we've built ourselves. This is of
> > course unfortunate.
> 
> Is this why builds of Clang have been failing for a while? I thought something terribly wrong happened, haha!

I think so. But I want to get a clean build environment before I can investigate more. I was out for two surgeries and wasn't able to take good care of the snapshots. Sorry that you had to notice this! And thank you for noticing it, I guess!

kwk:
>

PR #24, opened 2023-02-23
Build llvm-snapshot-builder from SRPM in today's Copr project

kwk:
In theory this change means that we can drop `packit` but I would like
to keep it in case things don't pan out as planned.

Packit right now checks PRs which is nice as well.


tstellar:
Don't we also need to hard-code the date into macros.llvm-snapshot-builder ?

kwk:
@tstellar yes. But like I mentioned on slack. This is yet to come. Right now, this is moving the package more local. Nothing more.

tstellar:
Looks good.

PR #28, opened 2023-03-02
llvm-snapshot-builder: Fix yyyymmdd definition

tstellar:
I don't think defines are allowed in the system macro files, so this was causing an error:

error: /usr/lib/rpm/macros.d/macros.llvm-snapshot-builder: line 209: Macro % has illegal name (%define)


kwk (approved):
LGTM. How great that an error doesn't cause a failure :/ Thank you for finding and fixing this. I have no way to try this but it looks good and since the evaluation is doing

```lua
if not rpm.expand('%yyyymmdd') ~= '%yyyymmdd' then
```

in the macros file, this shouldn't matter if it is a define or a regular macro. Fingers crossed.

PR #33, opened 2023-04-19
Re-enable automatic metadata generation for incubator projects

tstellar:
We need to be able to to install the llvm-snapshot-builder package from the current COPR project during the SRPM build phase.  This phase does not have access to to the internal repositories, so it needs to be able to install the llvm-snapshot-builder package from the public COPR repositories, which is only possible if we are re-generating the metadata aftereach build.

See https://src.fedoraproject.org/rpms/libomp/c/471201f1e41afcfab9b657a279d460e03792994e


kwk:
@tstellar this makes sense. I guess you only noticed it because your last PR #32 wasn't working the way you expected it and instead the llvm-snapshot-builder package was running on old version, right?

kwk (approved):
LGTM. Thanks

tstellar:
> @tstellar this makes sense. I guess you only noticed it because your last PR #32 wasn't working the way you expected it and instead the llvm-snapshot-builder package was running on old version, right?

Correct.

PR #34, opened 2023-05-04
Sleep after deleting copr project

nikic:
Per https://github.com/fedora-copr/copr/issues/2698#issuecomment-1534343813 there is a potential race condition between deleting and forking the project. Wait a minute between delete and fork to work around this for now.


kwk (approved):
LGTM

PR #36, opened 2023-05-22
Stop building for Fedora 36

nikic:
Fedora 36 has reached end-of-life on 2023-05-16, so stop providing snapshot builds for it.

Trying to produce snapshots for Fedora 36 currently breaks snapshots entirely, as building i386 for Fedora 36 on COPR is no longer supported.


kwk (approved):
Thank you @nikic for this PR. LGTM

PR #37, opened 2023-05-24
Build for latest fedora versions

kwk:
Automatically decide for which fedora versions to build based on `copr
list-chroots`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ yyyymmdd ?= $(today)
 buildroot ?= $(pwd)/buildroot/$(yyyymmdd)
 
 # The default chroot to build for.
-chroot ?= fedora-37-x86_64
+chroot ?= $(shell copr list-chroots | grep -P '^fedora-[0-9]+-x86_64' | sort | head -1)
 
 # The mock config that will be used as a template for the final mock config.
 # In it we will place the chroot defined above.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@ yyyymmdd ?= $(today)
 buildroot ?= $(pwd)/buildroot/$(yyyymmdd)
 
 # The default chroot to build for.
-chroot ?= fedora-37-x86_64
+chroot ?= $(shell copr list-chroots | grep -P '^fedora-[0-9]+-x86_64' | sort | head -1)
```
nikic, line 21: Maybe we should just use rawhide here?

nikic: I don't think so. I mean, we're building the snapshots on rawhide as well, and they're about as stable as the rest.

Probably the main disadvantage of rawhide builds is that you likely have to download more updated packages than if you build for something like f37.

nikic (approved):

nikic:
Looks like this broke the build (https://github.com/kwk/llvm-daily-fedora-rpms/actions/runs/5127777505/jobs/9223784866):
> Error: Unable to process file command 'env' successfully.
> Error: Invalid format 'fedora-37-i386'

nikic:
I think the problem may be that `chroots` contains newlines, which breaks the dump into `$GITHUB_ENV`.

PR #39, opened 2023-05-31
Avoid newlines in chroot variable

nikic:
The current way of exporting into $GITHUB_ENV only works for single-line values. Replace the newlines in the chroot list with spaces.

### DIFF
--- a/.github/workflows/fedora-copr-build.yml
+++ b/.github/workflows/fedora-copr-build.yml
@@ -33,7 +33,7 @@ jobs:
           yesterday=`date -d "${today} -1 day" +%Y%m%d`
 
           packages="python-lit llvm clang lld compiler-rt libomp"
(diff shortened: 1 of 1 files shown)


nikic:
https://medium.com/@ibraheemabukaff/github-actions-exporting-multi-line-one-line-value-environment-variable-5bb86d01e866 under "Environment Variable With Multi-Line Value" describes how to export a multi-line value, which looks kind of ugly to me, so I went with removing the newlines.

kwk (approved):
Thank you @nikic I was just going through my emails from this night and found that the [build](https://github.com/kwk/llvm-daily-fedora-rpms/actions/runs/5127777505/jobs/9223784866) had failed. Thank you for providing the solution. I've only fixed the position of the backtick.

PR #40, opened 2023-09-22
Automate wave through

nikic:
Automatically fork yesterday's snapshot incubator if all builds in it succeeded (as defined by the monitor page -- i.e. builds succeeding after resubmits is fine).

This matches the criterion for our current manual wave through, which has become something of a PITA (as builds only finish after working hours).

### DIFF
--- a/.github/workflows/fedora-copr-build.yml
+++ b/.github/workflows/fedora-copr-build.yml
@@ -51,14 +51,12 @@ jobs:
         run: |
           source github/functions.sh
 
-          # Check if yesterday's project exists and is waved through
+          # Check if yesterday's project exists and all builds succeeded
           yesterdays_project_exists=`project_exists ${{ env.project_yesterday }}`
           if [[ "$yesterdays_project_exists" == "true" ]]; then
-            url=https://copr.fedorainfracloud.org/coprs/`echo ${{ env.project_yesterday }} | sed -s 's/^@/\/g\//'`
-            curl -sL -o page.html $url 2>&1
-            grep "`head -n1 project-description.md`" page.html \
-              && yesterdays_project_exists="false" \
-              || true
+            if has_not_succeeded_builds ${{env.project_yesterday}}; then
+              yesterdays_project_exists=false
+            fi
           fi
 
           echo "todays_project_exists=`project_exists ${{ env.project_today }}`" >> $GITHUB_ENV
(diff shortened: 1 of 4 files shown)


kwk (commented):
@nikic thank you for your contribution on automatically waving through snapshot builds. I've took the freedom and extended your great work a bit: 45aa51496413d4d0aa3b420ab06c1c3e0344e939 Let me know if the rationale of the commit makes sense to you or needs explanation.

.github/workflows/fedora-copr-build.yml
```diff
@@ -44,40 +46,19 @@ jobs:
...
-          # Check if yesterday's project exists and is waved through
+          # Check if yesterday's project exists and all builds succeeded
           yesterdays_project_exists=`project_exists ${{ env.project_yesterday }}`
           if [[ "$yesterdays_project_exists" == "true" ]]; then
-            url=https://copr.fedorainfracloud.org/coprs/`echo ${{ env.project_yesterday }} | sed -s 's/^@/\/g\//'`
-            curl -sL -o page.html $url 2>&1
-            grep "`head -n1 project-description.md`" page.html \
-              && yesterdays_project_exists="false" \
-              || true
+            if has_not_succeeded_builds ${{env.project_yesterday}}; then
```
nikic, line 59: You've renamed the function, but the old name is still used here.

github/functions.sh
```diff
@@ -0,0 +1,57 @@
...
+# Returns false if a package needs special handling on certain architectures
+function is_package_supported_by_chroot() {
+  local pkg=$1;
+  local chroot=$2;
+
+  if [[ ("$pkg" == "lld" || "$pkg" == "libomp") && $chroot =~ -s390x$ ]]; then
+    false
+  else
+    true
+  fi
```
nikic, line 37: I think this can be reduced to just this?
```suggestion
  ! [[ ("$pkg" == "lld" || "$pkg" == "libomp") && $chroot =~ -s390x$ ]]
```

kwk: While it can be shortened, I dislike it.

.github/workflows/fedora-copr-build.yml
kwk: Thank you. Fixed in bee20b0.

nikic:
Looks good to me (but I can't approve my own PR ^^).

kwk (approved):
Let's give it a try.

nikic:
And of course it failed ... I believe https://github.com/kwk/llvm-daily-fedora-rpms/pull/41 should fix it.

PR #41, opened 2023-09-23
Run checkout action earlier

nikic:
The "variables and functions" step now makes use of github/functions.sh, so we should check out the repo before that.


kwk:
These three commits were necessary to fix everything:

a3a51b6e6a5c9f9515b0bcba6de79710d3bd92b6 (sorry that I didn't see your [this] PR first). 
ac94623391065b08d33307ffd70bfd163dcc4145
d65a306b268a58f9e1a7f1b038631e5eeabd4b28

PR #45, opened 2023-09-25
Fix --after-build-id if some builds are skipped

nikic:
If we skip some of the builds, we should set --after-build-id to the last non-skipped build. Otherwise we end up running the build too early (after the llvm build for a different chroot, rather than the s390x chroot), which I believe is the cause for the common s390x compiler-rt failures we see.

### DIFF
--- a/.github/workflows/fedora-copr-build.yml
+++ b/.github/workflows/fedora-copr-build.yml
@@ -157,9 +157,9 @@ jobs:
                   --chroot ${chroot} \
                   ${{ env.project_today }} \
                   | tee ${pkg}.log
-              fi
 
-              after_build_id="--after-build-id `cat ${pkg}.log | grep -Po 'Created builds: \K(\d+)'`"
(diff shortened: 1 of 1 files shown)


kwk:
@nikic I still try to understand what's going on here. I bet that we had some errors with `compiler-rt` because we were setting the `--after-build-id` to an empty grep? I mean the log file that we grep doesn't exist if we exclude the package. Is that what this is about?

nikic:
@kwk The log file will be left over from a previous chroot, so we will add a dependency on the package from a different chroot. So for example s390x compiler-rt could depend on ppc64le lld instead of depending on s390x llvm.

kwk (approved):

kwk:
> @kwk The log file will be left over from a previous chroot, so we will add a dependency on the package from a different chroot. So for example s390x compiler-rt could depend on ppc64le lld instead of depending on s390x llvm.

Ouch. Sorry for making this mistake in the first place.

PR #52, opened 2023-09-28
Include yyymmdd in NVR of llvm-snapshot-builder

kwk:
This adds the year month date to the NVR as well as the git revision,
just like in the snapshots.

To test, run these steps:

```console
$ cd llvm-snapshot-builder
$ make rpm
$ ls tmp/RPMS/noarch/
llvm-snapshot-builder-3.0.3~pre20230928.gdcb724d-1.fc38.noarch.rpm
```

Verify that your output contains the version like above or at least
similar.

Fixes: https://issues.redhat.com/browse/LLVM-43

### DIFF
--- a/llvm-snapshot-builder/Makefile
+++ b/llvm-snapshot-builder/Makefile
@@ -5,8 +5,9 @@ SHELL := /bin/bash
 PACKAGE=$(shell basename $(CURDIR))
 TMP=$(CURDIR)/tmp
 SPECFILE=$(PACKAGE).spec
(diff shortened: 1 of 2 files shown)


tuliom (approved):

kwk:
For some odd reason I can't seem to get the `~` to work in the NVR. For example, for the other snapshots we have versions like this: `libomp 18.0.0~pre20230919.g928564caa5de8b` but for the snapshot builder we have: `llvm-snapshot-builder 3.0.3pre20230929.g60bbb4f` which is missing the `~` right before the `pre`.

PR #54, opened 2023-10-05
Check today's snapshots every 60 minutes.

kwk:
If an issue was already reported for today's snapshot being broken, we don't report or even check for builds being broken.

### DIFF
--- a/.github/workflows/check-todays-snapshot.yml
+++ b/.github/workflows/check-todays-snapshot.yml
@@ -2,9 +2,9 @@ name: "Check today's snapshot"
 
 on:
   schedule:
-    # Everyday at 13:00 UTC
+    # At every 45th minute
     # See https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule
-    - cron: "00 13 * * *"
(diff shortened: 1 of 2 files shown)


nikic:
Is anything bad going to happen is that project doesn't exist at the time of the first check?

kwk:
> Is anything bad going to happen is that project doesn't exist at the time of the first check?

What project do you mean? The github project or the copr project? Because in the code of this PR we're only checking for the github project; and that better exist ;)

kwk:
My bad. Now I see what you mean. Because we run the check earlier than usual, the project in copr might not exist. That's what has change by this PR. Let me make some checks.

kwk:
@nikic I've added a check to not run the check for failed builds but I kept the times at which to check. We could have changed it to something like `*/45 8-23 * * *` which would translate to `At every 45th minute past every hour from 8 through 23.`. But I prefer it this way because I don't know exactly when builds are present or the next day's copr project is created. Checking for it is good enough.

nikic (dismissed):
This looks reasonable to me, though every 45 minutes is maybe a bit more than we really need?

PR #63, opened 2023-10-13
Add missing --repo argument

nikic:
Otherwise it defaults to the current git checkout, while `actions/checkout` apparently just downloads the contents instead of creating a git repo in this action.

### DIFF
--- a/github/functions.sh
+++ b/github/functions.sh
@@ -8,7 +8,8 @@ function yyyymmdd() {
 # Checks if there's an issue for a broken snapshot reported today
 function was_broken_snapshot_detected_today() {
   local d=`yyyymmdd`
-  gh issue list --label broken_snapshot_detected --state all | grep $d > /dev/null 2>&1
+  gh --repo ${GITHUB_REPOSITORY} issue list \
+    --label broken_snapshot_detected --state all | grep $d > /dev/null 2>&1
 }
 
 # Checks if a copr project exists


kwk:
> Otherwise it defaults to the current git checkout, which we don't actually check out in this action.

I'm not sure I understand this correctly. What action are you talking about? In the [`check-todays-snapshot.yml`](https://github.com/kwk/llvm-daily-fedora-rpms/blob/30b13e7ef3d62dc917ab4c75ebe239943cdaada5/.github/workflows/check-todays-snapshot.yml#L29) action we do check out this repository first:

```
- uses: actions/checkout@v3
```

before [calling was_broken_snapshot_detected_today](https://github.com/kwk/llvm-daily-fedora-rpms/blob/30b13e7ef3d62dc917ab4c75ebe239943cdaada5/.github/workflows/check-todays-snapshot.yml#L57).

nikic:
Good point. I've checked the log and it says:
```

Deleting the contents of '/__w/llvm-daily-fedora-rpms/llvm-daily-fedora-rpms'
The repository will be downloaded using the GitHub REST API
To create a local Git repository instead, add Git 2.18 or higher to the PATH
Downloading the archive
Writing archive to disk
Extracting the archive
```
So we "checkout" by downloading an archive, not a git repo.

kwk (approved):
Brilliant and well spotted. Thank you.

PR #73, opened 2023-10-24
Pass a new "snapshot" context to tmt

kwk:
This relates to
https://src.fedoraproject.org/tests/llvm/pull-request/19.

We don't built all LLVM packages as snapshot that are available in
Fedora. Therefore we must exclude certain packages from being installed.
When a LIT test runs that requires a package and it is not there, the
test will be unsupported and not cause errors.


jchecahi:
LGTM!

PR #74, opened 2023-10-25
tmt test bundle

kwk:
- tmt: add build-gating.fmf from clang
- Include tests for all snapshot components

### DIFF
--- a/tests/build-gating.fmf
+++ b/tests/build-gating.fmf
@@ -12,31 +12,37 @@ discover:
       how: fmf
       url: https://src.fedoraproject.org/tests/clang.git
       ref: main
-      filter: "tag:-not-in-default"
+      filter: "tag:-spoils-installation & tag:-not-in-default"
     - name: compiler-rt-tests
       how: fmf
       url: https://src.fedoraproject.org/tests/compiler-rt.git
       ref: main
+      filter: "tag:-spoils-installation & tag:-not-in-default"
     - name: libomp-tests
       how: fmf
       url: https://src.fedoraproject.org/tests/libomp.git
       ref: main
+      filter: "tag:-spoils-installation & tag:-not-in-default"
     - name: upstream-llvm-integration-testsuite
       how: fmf
       url: https://src.fedoraproject.org/tests/llvm.git
       ref: main
       test: integration-test-suite
+      filter: "tag:-spoils-installation & tag:-not-in-default"
     - name: python-lit
       how: fmf
       url: https://src.fedoraproject.org/tests/python-lit.git
       ref: main
+      filter: "tag:-spoils-installation & tag:-not-in-default"
     - name: lld-tests
       how: fmf
       url: https://src.fedoraproject.org/tests/lld.git
       ref: main
+      filter: "tag:-spoils-installation & tag:-not-in-default"
     - name: redhat-rpm-config-tests
       how: fmf
       url: https://src.fedoraproject.org/rpms/redhat-rpm-config.git
       ref: main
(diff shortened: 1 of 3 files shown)


kwk:
TODO: Adjust snapshot builder so it doesn't pick up the newly added plan.

tests/build-gating.fmf
```diff
@@ -0,0 +1,42 @@
...
+      ref: main
+      test: integration-test-suite
+    - name: python-lit
+      how: fmf
+      url: https://src.fedoraproject.org/tests/python-lit.git
+      ref: main
+    - name: lld-tests
+      how: fmf
+      url: https://src.fedoraproject.org/tests/lld.git
+      ref: main
```
jchecahi, line 36: Under this line add `filter: "tag:-spoils-installation & tag:-not-in-default"`. Else you plan will include [this test](https://src.fedoraproject.org/tests/lld/blob/main/f/ld-alternative), which will break your environment.
I would add that filter in each discover step, in case more breaking tests are added in the future.

tests/build-gating.fmf
```diff
@@ -0,0 +1,42 @@
+#
+# Snapshot gating tests for all *LLVM* packages in the mix
+#
+# Compatible with various these distributions:
+#
+#   * Fedora
+#
+
+summary: LLVM Tests for snapshot gating
+discover:
```
jchecahi, line 10: You could configure here the plan to automatically enable the copr repo you want to test from if you add the following block:
```
prepare:
  how: install
  copr: "@fedora-llvm-team/llvm-snapshots-incubator-$TODAY"
```
And then pass $TODAY from your workflow script like this:
```
tmt <context> run -e TODAY=$today <etc>
```

tests/build-gating.fmf
kwk: Done in https://github.com/kwk/llvm-daily-fedora-rpms/pull/74/commits/0c5a8d9a882b2d8686751bc16b98ff10018bc342

tests/build-gating.fmf
```diff
@@ -0,0 +1,48 @@
...
+      how: fmf
+      url: https://src.fedoraproject.org/tests/compiler-rt.git
+      ref: main
+      filter: "tag:-spoils-installation & tag:-not-in-default"
+    - name: libomp-tests
+      how: fmf
+      url: https://src.fedoraproject.org/tests/libomp.git
+      ref: main
+      filter: "tag:-spoils-installation & tag:-not-in-default"
+    - name: upstream-llvm-integration-testsuite
```
jchecahi, line 26: This discover config only gets integration-test-suite from LLVM tests. If you want to run all the LLVM tests I'd suggest to remove [this line](https://github.com/kwk/llvm-daily-fedora-rpms/pull/74/files#diff-48fcdeadae4f302e365808cdd08ee1cd02b46c59817c4269207104c7035cf935R30) (`test: integration-test-suite`), and maybe rename this discover config to "llvm-tests".

tests/build-gating.fmf
kwk: Awesome. Thank you! Done in [a146f40](https://github.com/kwk/llvm-daily-fedora-rpms/pull/74/commits/a146f40b36a5eb0c04d5a46285ec54e1d9af74c7)

tests/build-gating.fmf
kwk: I've seen this to be special but I wasn't sure (and frankly, didn't check) how the llvm repo tests this. But you're right. If I can test everything, then I want to do that. Done in 3592d34.

PR #78, opened 2023-10-25
tmt: exclude ppc64le-long-double

kwk:
Because it requires qemu and mock and we're running tmt inside a
container.

See https://src.fedoraproject.org/tests/clang/blob/main/f/ppc64le-long-double/main.fmf#_19

### DIFF
--- a/tests/snapshot-gating.fmf
+++ b/tests/snapshot-gating.fmf
@@ -26,6 +26,11 @@ discover:
       url: https://src.fedoraproject.org/tests/clang.git
       ref: main
       filter: "tag:-spoils-installation & tag:-not-in-default"
+      exclude:
+        # The following test only runs with qemu and we're executing in a github
+        # action runner in a container.
+        # See https://src.fedoraproject.org/tests/clang/blob/main/f/ppc64le-long-double/main.fmf#_19
+        - ppc64le-long-double
     - name: compiler-rt-tests
       how: fmf
       url: https://src.fedoraproject.org/tests/compiler-rt.git


kwk:
According to [this action run](https://github.com/kwk/llvm-daily-fedora-rpms/actions/runs/6646640885/job/18060516400#step:6:172), the test was successfully excluded. Notice the absence of `ppc64le-long-double` in the following list.

```
        execute task #1: clang-tests on default-0
        how: tmt
            00:00:00 pass /clang-tests/clang-format-diff (on default-0) [1/15]
            00:00:00 fail /clang-tests/fedora-flags (on default-0) [2/15]
            00:00:00 fail /clang-tests/gcc-clang-compatibility (on default-0) [3/15]
            00:00:00 fail /clang-tests/libomp (on default-0) [4/15]
            00:00:01 fail /clang-tests/llvm-test-suite (on default-0) [5/15]
            00:00:00 pass /clang-tests/long-double (on default-0) [6/15]
            00:00:02 fail /clang-tests/openmp-rpm (on default-0) [7/15]
            00:00:10 pass /clang-tests/rhbz_1647130 (on default-0) [8/15]
            00:00:00 fail /clang-tests/rhbz_1657544 (on default-0) [9/15]
            00:00:00 fail /clang-tests/rhbz_2239619 (on default-0) [10/15]
            00:00:00 fail /clang-tests/rhbz_482491 (on default-0) [11/15]
            00:00:00 pass /clang-tests/rpmmacros (on default-0) [12/15]
            00:00:07 fail /clang-tests/toolchains (on default-0) [13/15]
            00:00:00 pass /clang-tests/ucrt64-toolchain (on default-0) [14/15]
            00:00:00 fail /clang-tests/use-dwarf4-by-default (on default-0) [15/15]
```

tuliom:
@kwk The `ppc64le-long-double` already has code to detect if it should be ignored or not. There is something else going on in this execution that is causing the detection code to fail.
It would be great if we could fix the detection code instead of adding workarounds downstream.

kwk:
> @kwk The `ppc64le-long-double` already has code to detect if it should be ignored or not. There is something else going on in this execution that is causing the detection code to fail. It would be great if we could fix the detection code instead of adding workarounds downstream.

I get that. I think the ordering of [these two checks](https://src.fedoraproject.org/tests/clang/blob/main/f/ppc64le-long-double/main.fmf#_19) is causing the problem as the test is enabled if we're on fedora (check) on x86_64 (check) and not running ppc64le (check):

```
adjust:
  # ...
  - because: "Fedora CI runs in x86_64 only, emulate with qemu and mock"
    require+:
    - qemu-user-static
    - mock
    test: ./runtest-fedora.sh
    when: distro == fedora and arch == x86_64
    continue: false

  - enabled: false
    when: arch != ppc64le
```

The second `enabled: false` is not even reached because the first condition holds true. Maybe the [`adjust` section](https://fmf.readthedocs.io/en/latest/features.html#adjust) can have a check if it is running in a container or not; otherwise it will try to install mock and qemu. 

@jchecahi what do you think?

jchecahi:
> The second enabled: false is not even reached because the first condition holds true. 

@kwk I'm not sure to understand what you mean here. This is what happens in the `adjust` block:
* If `distro == fedora` and `arch == x86_64`, install qemu, mock and test with `runtest-fedora.sh` script. Do *NOT* continue making adjustments and run with this setup.
* Else, proceed with the rest of adjustments.
> Maybe the [adjust section](https://fmf.readthedocs.io/en/latest/features.html#adjust) can have a check if it is running in a container or not; otherwise it will try to install mock and qemu.

I'm not aware of any way to achieve that. `tmt` makes adjustments based on the context passed. Those adjustments do not necessarily happen in the system where the tests will run, in fact, so there's no way to guarantee that such kind of adjustments would be reliable.

PR #89, opened 2023-10-31
Set me as the current snapshot maintainer

tuliom:


nikic:
I think @kwk will also have to add you as a collaborator, otherwise the issue assignment is probably going to fail.

kwk (approved):

PR #91, opened 2023-11-03
libomp: build it on s390x

kwk:
See  https://github.com/llvm/llvm-project/pull/66081


nikic:
I think we first need to drop the `ExcludeArch: s390x` from the libomp spec file.

kwk:
@nikic you're right. Here's the PR for libomp: https://src.fedoraproject.org/rpms/libomp/pull-request/55

tuliom (approved):

PR #138, opened 2023-12-21
docs: change project description and instructions

kwk:
The new description should bring more clarity into what can be expected from the snapshots and how the organization with incubator projects is done.


tuliom (approved):

kwk:
The new description should bring more clarity into what can be expected from the snapshots and how the organization with incubator projects is done.

PR #162, opened 2024-01-02
github workflows: use matrix strategy on copr build and check-snapshots

kwk:
I found myself duplicating things again for the upcoming bootstrap
builds and figured it is best to employ a matrix strategy for
building snapshots and for checking them. That is easily extensible.
Just consider moving from big-merge to bootstrap and then to PGO.
All will benefit from this change.

Additionally I've added strategy labels (`strategy/standalone` and
`strategy/big-merge`) labels to issues that report a broken snapshot
build. This allows us to better filter for already existing issues.


.github/workflows/check-todays-snapshot.yml
```diff
@@ -10,6 +10,20 @@ on:
 
 jobs:
   check-todays-snapshot:
+    strategy:
+      matrix:
+        name: [standalone, big-merge]
+        include:
+          - name: standalone
+            maintainer_handle: "nickc"
```
nikic, line 18:
```suggestion
            maintainer_handle: "nikic"
```

nikic (approved):
Looks reasonable to me.

.github/workflows/check-todays-snapshot.yml
kwk: Good catch!

PR #173, opened 2024-01-08
Fix promotion check

nikic:
https://github.com/kwk/llvm-daily-fedora-rpms/commit/3119eda358a47007be7c8c50951e16550cee298f added a new parameter to has_all_good_builds, but only passes it in check-todays-snapshot.yml, but not fedora-copr-build.yml. For that reason, we always get a diff due to extra llvm-snapshot-builder lines and don't promote the successful snapshot build.


kwk (approved):
@nikic thank you for addressing my mistake!

PR #174, opened 2024-01-08
Automatically identify build failure causes

kwk:
This allows us to continously monitor a given day for error causes.

When a new issue is created for a broken snapshot build, we'll check the
logs of the failing builds for anything that we identify as an error.
The errors we can identify so far are listed below. A github issue label (i.e. `error/copr_timeout`, `error/network_issue`) is then automatically added to the broken snapshot issue.

# Examples of error causes

## `copr_timeout`

<pre>
 <b>!! Copr timeout</b> => sending INT
WARNING: Machine 5851d3319caf45608112626e5dff90dc still running. Killing...
Copr build error: Build failed
</pre>

## `network_issue`

<pre>
<b>Errors during downloading metadata for repository</b> 'local':
  - Status code: 404 for https://kojipkgs.fedoraproject.org/repos/rawhide/latest/x86_64/repodata/c3262644a11234a683575d44416933749f6808ea60694a499b11336c3b735223-filelists.xml.gz (IP: 38.145.60.20)
  - Status code: 404 for https://kojipkgs.fedoraproject.org/repos/rawhide/latest/x86_64/repodata/729723bef89105229ca2204cb1fdb395e04630c92e8119a2a3eec1c70dcb1a62-primary.xml.gz (IP: 38.145.60.20)
Error: Failed to download metadata for repo 'local': Yum repo downloading error: Downloading error(s): repodata/729723bef89105229ca2204cb1fdb395e04630c92e8119a2a3eec1c70dcb1a62-primary.xml.gz - Cannot download, all mirrors were already tried without success; repodata/c3262644a11234a683575d44416933749f6808ea60694a499b11336c3b735223-filelists.xml.gz - Cannot download, all mirrors were already tried without success
</pre>

## `test_issue`

<pre>
PASS: lit :: shtest-define.py (73079 of 73079)
********************
<b>Failed Tests</b> (1):
  libarcher :: races/task-taskgroup-unrelated.c

Testing Time: 226.85s

<b>Total Discovered Tests:</b> 96435
  Skipped          :    50 (0.05%)
  Unsupported      :  2416 (2.51%)
  Passed           : 93787 (97.25%)
  Expectedly Failed:   181 (0.19%)
  Failed           :     1 (0.00%)
</pre>

## `dependency_issue`

<pre>
No matching package to install: 'llvm-googletest = 18.0.0~pre20240108.g60c4f82d3c4e9c'
No matching package to install: 'llvm-static = 18.0.0~pre20240108.g60c4f82d3c4e9c'
No matching package to install: 'llvm-test = 18.0.0~pre20240108.g60c4f82d3c4e9c'
<b>Not all dependencies satisfied</b>
Error: Some packages could not be found.
WARNING: DNF command failed, retrying, attempt #2, sleeping 10s
No matches found for the following disable plugin patterns: local, spacewalk, versionlock
</pre>

## `unknown`

Anything that we cannot identify as one of the above error causes will be classified as `unknown`. Those are probably the most interesting errors to look into. That is why this change materializes the tail of the build log as a comment on the issue related to a broken snapshot.

### DIFF
--- a/.github/workflows/check-todays-snapshot.yml
+++ b/.github/workflows/check-todays-snapshot.yml
@@ -40,7 +40,7 @@ jobs:
 
       - name: Install Copr CLI
         run: |
-          dnf install -y copr-cli diffutils
+          dnf install -y copr-cli diffutils pcre2-tools
 
       - uses: actions/checkout@v3
 
@@ -74,16 +74,35 @@ jobs:
           source github/functions.sh
           [[ ! -z "${{ matrix.extra_script_file }}" ]] && source ${{ matrix.extra_script_file }}
 
-          if was_broken_snapshot_detected_today ${{ matrix.name }}; then
-            echo "We already filed an issue for broken snapshots today."
+          # This function inspects causes of build errors and adds a comment to
+          # today's issue.
+          function handle_error_causes() {
+            local causes_file=$1
+            local issue_number=`todays_issue_number ${GITHUB_REPOSITORY} ${{ matrix.name }}`
+            local error_labels=""
+            local comment_file=`mktemp`
+
             echo "Maybe we can identify new causes for errors by inspecting the build logs."
 
-            issue_number=`todays_issue_number ${{ matrix.name }}`
-            for error_cause in `list_error_causes ${{ env.project_today }}`; do
-              gh --repo ${GITHUB_REPOSITORY} label create error/$error_cause --force
-              gh --repo ${GITHUB_REPOSITORY} issue edit --add-label $issue_number error/$error_cause"
-            done
+            # If no error causes file was passed, process build logs to get
+            # error causes. Also ensure the error cause labels are created.
+            if [[ -n "$causes_file" && ! -f "$causes_file" ]]; then
+              causes_file=`mktemp`
+              error_causes=`list_error_causes ${{ env.project_today }} $causes_file`
+              create_labels_for_error_causes ${GITHUB_REPOSITORY} $error_causes
+            fi
 
+            # Turn some error causes into their own comment.
+            report_build_issues \
+              ${GITHUB_REPOSITORY} \
+              $issue_number \
+              $causes_file \
+              ${{ matrix.maintainer_handle }}
+          }
+
+          if was_broken_snapshot_detected_today ${GITHUB_REPOSITORY} ${{ matrix.name }}; then
+            echo "We already filed an issue for broken snapshots today."
+            handle_error_causes
             exit 0;
           fi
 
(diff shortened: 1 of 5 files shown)


nikic:
Nice idea! Some thoughts:

 * It would be good to also identify dependency failures, as these almost always occur as a result of other failures. Can probably look for "Not all dependencies satisfied". Example log: https://download.copr.fedorainfracloud.org/results/@fedora-llvm-team/llvm-snapshots-incubator-20240108/fedora-39-i386/06871097-clang/builder-live.log.gz
 * It would be good to also add a label for `error/unknown`, to indicate that there is at least one error that is "interesting" and should be looked into.
 * I think it would be good to also post a comment if a new "unknown" error is encountered. New labels aren't going to be noticed (as they don't send notifications). (We probably don't want to comment for network error / dependency error, as those don't really require action from our side.)

nikic:
> That's also a good idea. I have found a way to not re-post the unknown error on a per chroot+project basis. Therefore I think it is best to add an invisible HTML comment in the form `<!--error/$cause project/$project chroot/$chroot-->` to the comment body. Then I can check for the existence of such a comment more easily. For example, to find this very comment you can create this helper function:

What I had in mind here is to post a comment the first time the `error/unknown` label gets added, so the presence of the label could be used to check whether to post a comment.

If I understand your idea right, do you want to post a comment for each chroot that fails with an unknown cause? I think that would result in a lot of comments (probably about 15 if all llvm builds fail.) Maybe I understood what you have in mind though.

kwk:
> > That's also a good idea. I have found a way to not re-post the unknown error on a per chroot+project basis. Therefore I think it is best to add an invisible HTML comment in the form `<!--error/$cause project/$project chroot/$chroot-->` to the comment body. Then I can check for the existence of such a comment more easily. For example, to find this very comment you can create this helper function:
> 
> What I had in mind here is to post a comment the first time the `error/unknown` label gets added, so the presence of the label could be used to check whether to post a comment.

Well, it could be that different architecture fail for different unknown error. That is why I disliked this idea and haven't implemented it.

> If I understand your idea right, do you want to post a comment for each chroot that fails with an unknown cause? I think that would result in a lot of comments (probably about 15 if all llvm builds fail.) Maybe I understood what you have in mind though.

That is about right. We can tweak the mechanism by changing the HTML comment. If we change it to ignore the `chroot` argument, we have what you wanted.

See this for the matching part:

https://github.com/kwk/llvm-daily-fedora-rpms/pull/174/files#diff-2798660ab91227293c87361169e1048c324cea7b2f53f010487e41c216aafba8R131

And this for the creation part:

https://github.com/kwk/llvm-daily-fedora-rpms/pull/174/files#diff-2798660ab91227293c87361169e1048c324cea7b2f53f010487e41c216aafba8R243

Please note, that we're now able to materialize build log snippets into github comments like displayed in the following screenshot. Please note that I've used a network issue to demonstrate this.

grafik

I see great value in this because Copr removes build logs after a month or so. I'd love to see snippets like this test report in a github comment for example:

```
PASS: lit :: shtest-define.py (73079 of 73079)
********************
Failed Tests (1):
  libarcher :: races/task-taskgroup-unrelated.c


Testing Time: 226.85s

Total Discovered Tests: 96435
  Skipped          :    50 (0.05%)
  Unsupported      :  2416 (2.51%)
  Passed           : 93787 (97.25%)
  Expectedly Failed:   181 (0.19%)
  Failed           :     1 (0.00%)
```

PR #175, opened 2024-01-08
copr: add rhel-9-x86_64 build root as an experiment

kwk:
As discussed in today's meeting.

### DIFF
--- a/github/functions-big-merge.sh
+++ b/github/functions-big-merge.sh
@@ -1,6 +1,13 @@
 # This file contains overwrite for the functions in functions.sh
 set +x
 
+# Prefixes the old version of the given function with "_" to make it callable
+# from the overwriting function.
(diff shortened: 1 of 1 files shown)


github/functions-big-merge.sh
```diff
@@ -1,6 +1,13 @@
 # This file contains overwrite for the functions in functions.sh
 set +x
 
+# Prefixes the old version of the given function with "_" to make it callable
+# from the overwriting function.
+function overwrite_function() {
+  local function_name=$1
+  eval "`declare -f $function_name | sed '1s/.*/_&/'`"
```
nikic, line 8:
```suggestion
  eval "_`declare -f $function_name`"
```
Would this work? I'm not really clear on what that `sed` call does.

nikic (dismissed):

PR #178, opened 2024-01-09
Set label colors

nikic:
Set label colors for arch/project/os/strategy labels. I set the colors manually yesterday so we have one color for reach label kind, but apparently they get re-set every time the labels are re-created. This means we basically get a new random label color every day :)

Fix this by explicitly passing `--color`. The chosen colors are from GitHub's standard set in the light variant.


kwk (approved):
LGTM. Thank you for addressing this and a per-catergory color makes a lot of sense.

kwk:
@nikic Please note, that only those labels get a defined color that failed.

PR #320, opened 2024-03-11
Filter snapshot commit selection based on list of checks only

kwk:
We now select a commit for the next snapshot based on a list of checks that must have succeeded for that commit. The combined commit status will be ignored as discussed with the team.


nikic (approved):
Looks reasonable from a cursory look.

PR #380, opened 2024-04-03
Invalidate old testing-farm request IDs and split up chroot comments

kwk:
This PR does many things.

* Invalidate old testing-farm request IDs if the Copr build IDs have changed.
  * Still allow old testing-farm HTML comments without any Copr build IDs and update them on the fly.
* Add testing-farm tests
* Exclude testing-farm-logs dir from pre-commit check
* Flip_test_label improvement and tests
* Fix error with authentication
* Fix Incomplete URL substring sanitization
* Render errors for each chroot in their own updatetable and hidable comment with a link in 
the chroot column of the build matrix.
  * Added ability to hide and unhide (minimize/unminimize) comments through GraphQL
  * Hide/Unhide all unused/used chroot error comments if they belong to a chroot we don't care about
* Fix rendering of test output
* Reorganize testing-farm bits
* Fix type hints in python code
* Shorten testing-farm log


kwk:
Consider the chroot fedora-38-x86_64 only having successful builds. What happens is, that we kick of testing farm. But then you think that you need to change something and you kick of a new build in the fedora-38-x86_64 chroot again and it succeeds. In that case the testing-farm output will stay which is wrong by design. I not only need to log the chroot and request ID of testing farm but I need to log the old build IDs with it in order to properly invalidate any testing-farm output.

snapshot_manager/snapshot_manager/testing_farm_util.py
```diff
@@ -20,6 +22,399 @@
...
 
+@dataclasses.dataclass(kw_only=True, unsafe_hash=True)
+class TestingFarmRequest:
+    request_id: uuid.UUID
+    """The request ID returned from a call to testing-farm request' """
+
+    chroot: str
+    """The chroot for which this testing-farm request was issues"""
+
+    copr_build_ids: list[int]
```
kwk, line 33: > Consider the chroot fedora-38-x86_64 only having successful builds. What happens is, that we kick of testing farm. But then you think that you need to change something and you kick of a new build in the fedora-38-x86_64 chroot again and it succeeds. In that case the testing-farm output will stay which is wrong by design. I not only need to log the chroot and request ID of testing farm but I need to log the old build IDs with it in order to properly invalidate any testing-farm output.

The Copr build IDs are now associated with a `TestingFarmRequest` object.

https://github.com/fedora-llvm-team/llvm-snapshots/pull/380/files#diff-5e0a4e83cc977056949528e7d51a250ad2187de6f6864bb8e3b9bfa690cc900dR33-R35

Issue #462, opened 2024-05-02
Make "big-merge" the default


kwk:
- [x] Disable standalone builds
- [x] Continue building big-merge into llvm-snapshots-big-merge-YYYYMMDD on copr but fork them into llvm-snapshots
- [x] Update copr project documentation about the new name
- [x] Check if we need to adapt build time diagrams
- [x] Move code from https://src.fedoraproject.org/fork/kkleine/rpms/llvm/tree/big-merge to https://src.fedoraproject.org/rpms/llvm/tree/big-merge (I moved my commits around and squashed them. You MUST rebase your patches)
- [x] Get rid of llvm-snapshot-builder
- [ ] Get documentation in README.adoc in good shape 
- [x] Adjust branch protection rules that no packit builds for llvm-snapshot-builder are required 
![grafik](https://github.com/fedora-llvm-team/llvm-snapshots/assets/193408/457ee1c5-e761-4eeb-986a-c33aa6eee7a6)
- [x] Decide what to do with `lld`. It was packaged in standalone mode but not in `big-merge`. Options are to continue building it standalone or merge it.

kwk:
Documentation can be done in a different task.

PR #466, opened 2024-05-03
fix 462

kwk:
- **Disable building standalone builds. See #462**
- **Disable checking standalone builds. See #462**
- **Fork big-merge copr builds into @fedora-llvm-team/llvm-snapshots. See #462**
- **Get rid of llvm-snapshot-builder. See #462**
- **Update copr project documentation about new name for incubator projects. See #462**
- **Make big-merge the default strategy in config. See #462**


.github/workflows/fedora-copr-build.yml
```diff
@@ -13,28 +13,25 @@ jobs:
...
+          #   copr_project_tpl: "@fedora-llvm-team/llvm-snapshots-incubator-YYYYMMDD"
+          #   copr_target_project: "@fedora-llvm-team/llvm-snapshots"
+          #   clone_url_tpl: "https://src.fedoraproject.org/rpms/PKG.git"
+          #   clone_ref: upstream-snapshot
           - name: big-merge
             copr_project_tpl: "@fedora-llvm-team/llvm-snapshots-big-merge-YYYYMMDD"
-            copr_target_project: "@fedora-llvm-team/llvm-snapshots-big-merge"
+            copr_target_project: "@fedora-llvm-team/llvm-snapshots"
             extra_script_file: "scripts/functions-big-merge.sh"
             clone_url_tpl: "https://src.fedoraproject.org/forks/kkleine/rpms/PKG.git"
```
nikic, line 27:
```suggestion
            clone_url_tpl: "https://src.fedoraproject.org/rpms/PKG.git"
```
I think this should point to the main repo now?

kwk: That is correct. Let me fix it.

nikic (approved):
LGTM if we're okay with temporarily publishing to llvm-snapshots without lld.

If not, then I think we first need to only do the change of the repo, then add lld support, make sure that it builds, and only then start promoting from the big-merge branch. Or alternatively build lld separately in the big-merge incubator.

nikic:
I've put up https://github.com/fedora-llvm-team/llvm-snapshots/pull/473 with just the repo change. I'd like to get the big-merge branch back to working ASAP without blocking it on the lld work.

PR #556, opened 2024-06-19
Set correct url for llvm tmt tests repo

jchecahi:
It was set to a different repo while some issues were fixed, now it's safe to go back to the upstream.


nikic (approved):

kwk:
Thank you @jchecahi this is really good!

PR #577, opened 2024-07-03
Add script to compare package contents

kwk:
The script checks the actual package contents of rawhide and the currently promoted snapshot build. It uses the systems architecture in order to run two containers with the right set of LLVM versions installed.

### DIFF
--- a/scripts/compare-package-contents.sh
+++ b/scripts/compare-package-contents.sh
@@ -16,7 +16,7 @@
 #           ├── ...
 #           ├── clang-libs.txt
 #
-# The "old" and "new" direcories refer to "rawhide" and "snapshots".
+# The "old" and "new" directories refer to "rawhide" and "snapshots".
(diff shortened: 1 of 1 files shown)


nikic (approved):
Thanks, works great!

I've reviewed the results, and found the following differences:

 * clang-libs ships way too many libraries. Fixed in https://src.fedoraproject.org/rpms/llvm/pull-request/241.
 * rawhide ships the clang cfg files in clang-libs, while big-merge ships them in clang. I think the big-merge behavior is correct, but maybe I'm missing some subtlety with library users of clang here. @tstellar Do you know why these config files were in clang-libs?
 * python3-clang in big-merge is just an empty directory! Something is wrong here.

nikic:
https://src.fedoraproject.org/rpms/llvm/pull-request/242 is the fix for the python-clang issue.

PR #595, opened 2024-07-18
Allow specifying the project in compare-package-contents.sh

nikic:
This allows using something like

    ./compare-package-contents.sh @fedora-llvm-team/llvm-snapshots-big-merge-20240718

to compare an incubator project rather than the promoted project.


kwk:
Have you tried running the script three time?

1. with the default
2. with a custom project
3. with the default

I wonder if the 3rd time around, the container cache is used to speed things up.

Otherwise this looks functionally correct to me. Just the issue with performance.

PR #598, opened 2024-07-19
rhel 8 support

kwk:
- **Enable RHEL 8 in builds**
- **Enable swig:4.0 module in RHEL 8 build chroots (if any)**

### DIFF
--- a/.github/workflows/fedora-copr-build.yml
+++ b/.github/workflows/fedora-copr-build.yml
@@ -136,7 +136,7 @@ jobs:
           [[ ! -z "${{ matrix.extra_script_file }}" ]] && source ${{ matrix.extra_script_file }}
 
           for chroot in ${{ env.chroots }}; do
-            if [[ "$chroot" ==  rhel-8- ]]; then
+            if [[ "$chroot" == rhel-8-* ]]; then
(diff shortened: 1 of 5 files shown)


.github/workflows/fedora-copr-build.yml
```diff
@@ -129,6 +129,18 @@ jobs:
...
             $chroot_opts "${{ env.project_today }}"
 
+      - name: "Enable swig:4.0 module in RHEL 8 build chroots (if any)"
+        shell: bash -e {0}
+        run: |
+          source scripts/functions.sh
+          [[ ! -z "${{ matrix.extra_script_file }}" ]] && source ${{ matrix.extra_script_file }}
+
+          for chroot in ${{ env.chroots }}; do
+            if [[ "$chroot" ==  rhel-8- ]]; then
```
nikic, line 139:
```suggestion
            if [[ "$chroot" == rhel-8-* ]]; then
```
I think?

kwk: Totally. In fact, see my console for my worst tries today. I had this but forgot to write it down properly. Thank you for finding this.

grafik

nikic (approved):

PR #618, opened 2024-07-31
attempt to fix testing farm watch timeout

kwk:
- **Declutter log output**
- **Bump timeout for testing-farm watch**

### DIFF
--- a/snapshot_manager/main.py
+++ b/snapshot_manager/main.py
@@ -12,7 +12,7 @@ def main():
 
     logging.basicConfig(
         level=logging.INFO,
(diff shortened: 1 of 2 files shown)


tuliom:
It looks like this continues to happen. I wonder if these requests are being blocked somewhere.

kwk:
> It looks like this continues to happen. I wonder if these requests are being blocked somewhere.

Yes they are. Meanwhile testing-farm was updated and they now include more details in the `watch` output which requires to make calls to each ranch. Obviously we cannot make these calls to the internal redhat ranch from within github. But testing-farm still tries it anyways. The testing-farm folks are giving us an option to disable the additional output and fetching from the ranches. Then we should be back on track.

PR #630, opened 2024-08-07
Reduce window for finding a good commit

### LINKED ISSUE #628: Reduce window for finding a good commit
Currently, we allow going back up to 500 commits to find a good one. This means that we end up building the same commit multiple times if buildbot is down, which is not particularly useful. This happened in https://github.com/fedora-llvm-team/llvm-snapshots/issues/627 for example.

We should reduce the window to a smaller value like 100 commits and return the most recent commit if we can't find a good one. I think it's better to test a new commit that's not known good rather than an old one we already tested.

kwk:
- Only check last 100 commits.
- Return the initial commit if no good one was found.

Fixes #628

### DIFF
--- a/scripts/get-good-commit.py
+++ b/scripts/get-good-commit.py
@@ -63,9 +63,7 @@ Required checks: {required_checks}
         return commit.sha
 
     sha = repo.get_commit(sha=start_ref).sha
-    logging.info(
-        f"No good commit found, using the initial one: {start_ref}, aka {sha}"
-    )
+    logging.info(f"No good commit found, using the initial one: {start_ref}, aka {sha}")
     return sha
 
 
--- a/scripts/get-good-commit.py
+++ b/scripts/get-good-commit.py
@@ -64,7 +64,7 @@ Required checks: {required_checks}
 
     sha = repo.get_commit(sha=start_ref).sha
     logging.info(
-        f"Not found a good commit, using the intial one: {start_ref}, aka {sha}"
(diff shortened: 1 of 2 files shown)


scripts/get-good-commit.py
```diff
@@ -62,12 +62,16 @@ def get_good_commit(
         logging.info(f"Found good commit: {commit_url}")
         return commit.sha
 
-    return ""
+    sha = repo.get_commit(sha=start_ref).sha
+    logging.info(
+        f"Not found a good commit, using the intial one: {start_ref}, aka {sha}"
```
nikic, line 67:
```suggestion
        f"No good commit found, using the initial one: {start_ref}, aka {sha}"
```

scripts/get-good-commit.py
```diff
@@ -62,12 +62,16 @@ def get_good_commit(
...
+    )
+    return {sha}
```
nikic, line 69:
```suggestion
    return sha
```
Or is this some weird new python syntax I don't know about?

kwk: No you're right. I don't know why I've used that. Probably because I starred at the f-strings.

nikic (approved):
LGTM

nikic:
Note that there is a python formatting failure.

kwk:
> Note that there is a python formatting failure.

Probably introduced by just accepting your suggestion in github. Let me see.

PR #631, opened 2024-08-07
Change how build failure comments are generated

### LINKED ISSUE #626: Change how build failure comments are generated
Currently each build failure produces a new comment. These comments come in over time. An unfortunate side effect of this is that the comments left by humans are hard to spot, because they may be followed by multiple automated comments.

I try to post any comments on snapshot build issues only once most builds have completed for this reason, but this is not really great...

A possible alternative would be to post dummy comments for all builds first and mark them as hidden, and then unhide them if there is an actual build failure.

When doing that, we should also only set the assignee after posting these comments. That way the maintainer will only receive one notification per day, rather than one per failed build.

kwk:
1. Upon issue creation we create minimized comments for all chroots and
   mark them with i.e. `<!--ERRORS_FOR_CHROOT/fedora-rawhide-x86_64-->`.
2. When we find an error in a chroot we update the above created comment
   and unminimize it.
3. In case when there's no error for a chroot, we're gonna minimize the
   comment again.

Fixes #626

### DIFF
--- a/snapshot_manager/snapshot_manager/github_util.py
+++ b/snapshot_manager/snapshot_manager/github_util.py
@@ -168,7 +168,7 @@ remove the aforementioned labels.
         if maintainer_handle is not None:
             opts["assignee"] = maintainer_handle
         issue = repo.create_issue(
-            title=self.issue_title(), body=self.initial_comment, *opts
+            title=self.issue_title(), body=self.initial_comment, **opts
         )
         self.create_labels_for_strategies(labels=[strategy])
         issue.add_to_labels(f"strategy/{strategy}")
--- a/snapshot_manager/snapshot_manager/github_util.py
+++ b/snapshot_manager/snapshot_manager/github_util.py
@@ -106,8 +106,6 @@ class GithubClient:
         llvm_release = util.get_release_for_yyyymmdd(self.config.yyyymmdd)
         llvm_git_revision = util.get_git_revision_for_yyyymmdd(self.config.yyyymmdd)
         return f"""
-Hello @{self.config.maintainer_handle}!
-
 <p>
 This issue exists to let you know that we are about to monitor the builds
 of the LLVM (v{llvm_release}, <a href="https://github.com/llvm/llvm-project/commit/{llvm_git_revision}">llvm/llvm-project@ {llvm_git_revision[:7]}</a>) snapshot for <a href="{self.config.copr_monitor_url}">{self.config.yyyymmdd}</a>.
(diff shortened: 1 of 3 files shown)


kwk:
> It would be good to address this part of the issue at the same time:
> 
> > When doing that, we should also only set the assignee after posting these comments. That way the maintainer will only receive one notification per day, rather than one per failed build.
> 
> Otherwise whoever is maintainer will get ~20 not very helpful emails for empty comments (unless GitHub suppresses them in this case?)

@nikic I've made the change as you suggested. We'll have to ask @tuliom tomorrow if he got more than one notification.

nikic:
Thanks! I think an adjustment to initial_comment may also be needed, as that also uses maintainer_handle ping (and thus subscribes by mail to the issue). I think it's fine to just drop the "hi" there :)

snapshot_manager/snapshot_manager/github_util.py
```diff
@@ -164,11 +161,8 @@ def create_or_get_todays_github_issue(
         repo = self.gh_repo
         logging.info("Creating issue for today")
 
-        issue = repo.create_issue(
-            assignee=maintainer_handle,
-            title=self.issue_title(),
-            body=self.initial_comment,
-        )
+        opts = {}
```
nikic, line 164: Unused opts?

snapshot_manager/snapshot_manager/snapshot_manager.py
```diff
@@ -176,6 +175,17 @@ def check_todays_builds(self) -> None:
 
         all_chroots = self.copr.get_copr_chroots()
 
+        if issue_is_newly_created:
+            # The issue was newly created so we'll create comments for each
+            # chroot that we care about and hide them for now. Then humanly
+            # created output will be come at the end always.
```
nikic, line 181:
```suggestion
            # created output will always come at the end.
```

nikic (approved):
LGTM, thanks!

snapshot_manager/snapshot_manager/github_util.py
kwk: Ups.

nikic:
I got 14 empty notifications todays :) It seems like there is a bit of a race condition on GitHub's side, so if a lot of actions are taken in a short time frame, it may end up sending notifications for stuff that happened before you subscribed to the issue.

This can probably be "fixed" just by waiting for a minute before doing the assignment...

kwk:
> I got 14 empty notifications todays :) It seems like there is a bit of a race condition on GitHub's side, so if a lot of actions are taken in a short time frame, it may end up sending notifications for stuff that happened before you subscribed to the issue.

I feared this because Github certainly does so when assigning and removing labels. It groups the actions together it seemed.

> This can probably be "fixed" just by waiting for a minute before doing the assignment...

How about assigning the maintainer only on the second run of the checks like done in https://github.com/fedora-llvm-team/llvm-snapshots/pull/643? The second run is still way to early to expect any results anyways.

PR #633, opened 2024-08-07
Downgrade testing-farm to 0.0.16

kwk:
This should allow us to get artifact URLs pointing to testing-farm results on the redhat ranch.

In 0.0.18 or before, testing-farm introduced more detailed output in the
`testing-farm watch` command which required to make calls that at least
the redhat ranch were only possible in the Red Hat VPN which is not
possible on github.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -85,10 +85,6 @@ requests-toolbelt==1.0.0
     # via copr
 rich==12.6.0
     # via typer
-ruamel-yaml==0.18.6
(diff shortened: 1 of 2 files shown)


nikic (approved):
LGTM

kwk:
I've created this upstream issue to understand if we will be able to update to anything above 0.0.16: https://gitlab.com/testing-farm/cli/-/issues/31.

Issue #671, opened 2024-08-26
Copr packages not installed in Fedora 41 and Rawhide


jchecahi:
In the testing-farm runs of fedora-41 and fedora-rawhide, tmt is installing LLVM 18 packages coming from the system repos rather than LLVM 20 packages coming from the copr repo. Examples from snapshot 20240825:
* [fedora-rawhide-x86_64](https://artifacts.dev.testing-farm.io/5987a921-7687-467a-b2b3-09378d3042b8/work-snapshot-gatingpx44tv8z/log.txt)
* [fedora-41-x86_64](https://artifacts.dev.testing-farm.io/8dbfef22-76b0-4968-a41b-ff687daa8198/work-snapshot-gatingojrombe_/log.txt)

jchecahi:
@kwk It turns out this issue is because of the testing-farm-tag repo priority, related to #395. In our test plan we decrease the priority of the repo in [this line](https://github.com/fedora-llvm-team/llvm-snapshots/blob/b21d85c41f5b44b1b0cdf5545f915566272ff30b/tests/snapshot-gating.fmf#L20). However, we call `dnf config-manager` using the old dnf4 interface, which `dnf5`, default package manager for Fedora 41 and newer, don't understand . Since we're piping the output to  a `true` command, the adjustment fails unadvertently, the priority change is not done, but tmt's prepare stage continues.
```
            dnf config-manager --save --setopt="testing-farm-tag-repository.priority=999" || true
[ cropped output, not relevant here ]
10:59:00             cmd: /var/ARTIFACTS/work-snapshot-gatingzljcusxv/tests/snapshot-gating/tree/tmt-prepare-wrapper.sh-default-1-default-0
10:59:00         environment
10:59:00             out: Updating and loading repositories:
[ cropped output, not relevant here ]
10:59:05             out: Complete!
10:59:05             err: Unknown argument "--save" for command "config-manager". Add "--help" for more information about the arguments.
```

I will work on a PR to solve this issue.

nikic:
I wonder whether https://github.com/fedora-llvm-team/llvm-snapshots/blob/main/tests/snapshot-gating.fmf could inject an additional test that just checks that `clang -v` (or similar) returns the expected version for snapshots...

kwk:
[...]

PR #676, opened 2024-08-27
Check that rpms installed in tmt come from copr

kwk:
Add a script to ensure that the llvm package installed comes from the copr repo. A failure in the script aborts the test plan

See #671

### DIFF
--- a/tests/snapshot-gating.fmf
+++ b/tests/snapshot-gating.fmf
@@ -23,6 +23,11 @@ prepare:
     script: |
       dnf install -y dnf-plugins-core
       dnf config-manager --save --setopt="testing-farm-tag-repository.priority=999" || true
+
+  - name: "Check that snapshot (~pre) version of LLVM is installed"
+    how: shell
+    order: 99
+    script: rpm -q --qf "%{version}" llvm-libs | grep '~pre'
 adjust:
   - discover+:
       - name: redhat-rpm-config-tests
--- a/tests/snapshot-gating.fmf
+++ b/tests/snapshot-gating.fmf
@@ -25,8 +25,11 @@ prepare:
   # See: https://docs.testing-farm.io/Testing%20Farm/0.1/test-environment.html#_tag_repository
   - how: shell
     script: |
-      dnf install -y dnf-plugins-core
-      dnf config-manager --save --setopt="testing-farm-tag-repository.priority=999" || true
+      if dnf repolist | grep -q testing-farm-tag-repository; then
+        dnf install -y 'dnf5-command(config-manager)' || dnf install -y 'dnf-command(config-manager)'
+        dnf config-manager --save --setopt="testing-farm-tag-repository.priority=999" || \
+          dnf config-manager setopt "testing-farm-tag-repository.priority=999"
+      fi
 
   - name: "Check that snapshot (~pre) version of LLVM is installed"
     how: shell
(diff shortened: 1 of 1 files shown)


tests/snapshot-gating.fmf
```diff
@@ -23,6 +23,11 @@ prepare:
     script: |
       dnf install -y dnf-plugins-core
       dnf config-manager --save --setopt="testing-farm-tag-repository.priority=999" || true
+
+  - name: Assert correct llvm package is installed
+    how: shell
+    order: 99
+    script: tests/assert_llvm.sh
```
kwk, line 30:
```suggestion
    script: rpm -qi llvm-libs | grep -P "Version.*~pre.*"
```

This should do the trick.

tests/assert_llvm.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
kwk, line 1: I think you can remove this file.

tests/snapshot-gating.fmf
jchecahi: I understand that your point here is to keep it simple. I do think that the script might be more beneficial in this case, though. The script gives us some advantages here:

- We can make several checks at the `prepare` step end, all grouped under the same script. This may include, checking that the copr repo is reachable.
- We can write more complex logic. For example testing several packages rather that llvm-libs only.
- We can properly report what failed and why.

For example, upon wrong package installation with the script, this is what we see in the report:
```
    prepare step failed

    The exception was caused by 1 earlier exceptions

    Cause number 1:

        Command '/var/ARTIFACTS/work-snapshot-gatingejv1c0ks/tests/snapshot-gating/tree/tmt-prepare-wrapper.sh-Assert-correct-llvm-package-is-installed-default-0' returned 1.

        stdout (2 lines)
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        *** Unexpected llvm package installed: llvm-18.1.8-2.fc41.x86_64
        *** Expected llvm: llvm-20.0.0~pre20240825.g3ef64f7ab5b865-1.fc42.x86_64
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
However, using the plain command, we would see this instead, which is much less intuitive:

```
    prepare step failed

    The exception was caused by 1 earlier exceptions

    Cause number 1:

        Command '/var/tmp/tmt/run-077/tests/snapshot-gating/tree/tmt-prepare-wrapper.sh-Assert-correct-llvm-package-is-installed-default-0' returned 1.

```

tests/snapshot-gating.fmf
```diff
@@ -23,6 +23,11 @@ prepare:
     script: |
       dnf install -y dnf-plugins-core
       dnf config-manager --save --setopt="testing-farm-tag-repository.priority=999" || true
+
+  - name: Assert correct llvm package is installed
```
kwk, line 27:
```suggestion
  - name: "Check that snapshot (~pre) version of LLVM is installed"
```

kwk:
Closed in favor of https://github.com/fedora-llvm-team/llvm-snapshots/pull/680

PR #697, opened 2024-09-04
Enforce installing llvm-libs from the copr repo

jchecahi:
This addresses test run issues raised in #692.

Tested on [this run](https://artifacts.osci.redhat.com/testing-farm/57fd8eab-d1a5-4e65-86ba-ad8405884e92/) (RedHat internal only).

### DIFF
--- a/tests/snapshot-gating.fmf
+++ b/tests/snapshot-gating.fmf
@@ -20,6 +20,10 @@ prepare:
       # parsing ourselves.
       # Check https://github.com/fedora-copr/copr/issues/3387
       - sed -i "s/\$distname/$(echo "$COPR_CHROOT" | cut -d '-' -f 1)/" /etc/yum.repos.d/*$COPR_PROJECT*
+      # Ensure that latest llvm-libs (from our snapshot) is installed to prevent
+      # potential dependency solving issues when rpms depending on llvm-libs are
+      # already in the system.
+      - dnf -y install --best llvm-libs
 
   # Lower the priority of the testing-farm-tag-repository so that our copr repo is picked up.
   # See: https://docs.testing-farm.io/Testing%20Farm/0.1/test-environment.html#_tag_repository


kwk (commented):
@jchecahi it looks good and seems to work. But I now wonder if we should ask dnf to install a specific version of our packages. We know for each day which version that is by asking:

```bash
llvm_snapshot_yyyymmdd=$(date +%Y%m%d)
git_revision_url=https://github.com/fedora-llvm-team/llvm-snapshots/releases/download/snapshot-version-sync/llvm-git-revision-${llvm_snapshot_yyyymmdd}.txt
llvm_snapshot_git_revision=$(curl -sL $git_revision_url)
llvm_snapshot_git_revision_short=$(echo "${llvm_snapshot_git_revision:0:14}")

release_url=https://github.com/fedora-llvm-team/llvm-snapshots/releases/download/snapshot-version-sync/llvm-release-${llvm_snapshot_yyyymmdd}.txt
llvm_snapshot_version=$(curl -sL $release_url)
llvm_snapshot_version_suffix=pre${llvm_snapshot_yyyymmdd}.g${llvm_snapshot_git_revision_short}
llvm_snapshot_version_tag=${llvm_snapshot_version}~${llvm_snapshot_version_suffix}

echo "Need to install llvm-libs-${llvm_snapshot_version_tag}"
```

I approve this PR but if you say that we can install a specific version then I'd like that better. I understand that we're pulling in tests that should be left untouched and are used for other OSes where we don't know the LLVM version upfront. Just saying.

jchecahi:
> 
> I approve this PR but if you say that we can install a specific version then I'd like that better. I understand that we're pulling in tests that should be left untouched and are used for other OSes where we don't know the LLVM version upfront. Just saying.

In #671 we agreed to keep this as simple as possible, and I think we should follow the same approach here. IMHO, it's not worth adding that complexity to the prepare step of a tmt test plan. 
We don't need to be that exhaustive and request the exact version, as we can make some safe assumptions in our environment:
1. We're adding a copr repo on top of a distro (be it rhel, be it fedora) which contains the snapshot for a given day. It's guaranteed that this copr repo (regardless the day) has an llvm newer than any distro repos.
2. We ask dnf to install the "best" llvm. From previous point, dnf will install it (or upgrade to) from the copr repo.
3. Let tmt install all the test requirements
4. Verify that llvm-libs installed is a `~pre` package, ensuring that tmt did not downgrade or mess up with our snapshot package.

PR #718, opened 2024-09-12
Add support for centos-stream-10 chroots to pgo strategy

kwk:
This adds centos-stream-10 chroots to the pgo strategy builds.


nikic:
Why only for the pgo strategy?

PR #732, opened 2024-09-17
Install llvm-test-suite from copr

nikic:
This enables the https://copr.fedorainfracloud.org/coprs/g/fedora-llvm-team/llvm-test-suite/monitor/ copr repo, so use a newer llvm-test-suite in the tmt tests.

### DIFF
--- a/tests/snapshot-gating.fmf
+++ b/tests/snapshot-gating.fmf
@@ -35,6 +35,8 @@ prepare:
       # potential dependency solving issues when rpms depending on llvm-libs are
       # already in the system.
       - dnf -y install --best llvm-libs
+      # Use a newer llvm-test-suite version from copr.
+      - dnf copr enable -y @fedora-llvm-team/llvm-test-suite $COPR_CHROOT
 
   - name: "Check that snapshot (~pre) version of LLVM is installed"
     how: shell


tests/snapshot-gating.fmf
```diff
@@ -36,6 +36,9 @@ prepare:
       # already in the system.
       - dnf -y install --best llvm-libs
 
+      # Use a newer llvm-test-suite version from copr.
```
jchecahi, line 39: This change makes sense to me. Just a little comment: could you remove the blank line above this one so we have the whole section packed?

tests/snapshot-gating.fmf
```diff
@@ -36,6 +36,9 @@ prepare:
...
+      - dnf copr enable -y @fedora-llvm-team/llvm-test-suite $COPR_CHROOT
```
kwk, line 40: @jchecahi I think we have another reason for lowering the testing-farm repo instead of raising our own: We now have two repos to add ;)

@nikic This looks good to me. But I wonder what will happen if we for example add a new chroot to the snapshot builds, we would need to add it here as well, right? For example, `centos-stream-10-...` is missing from the list in https://copr.fedorainfracloud.org/coprs/g/fedora-llvm-team/llvm-test-suite/monitor/.

tests/snapshot-gating.fmf
nikic: Done!

tests/snapshot-gating.fmf
nikic: Yes, we'd have to add the new chroots here, similar to how we do in llvm-compat-packages. Though we don't need to do anything for new fedora versions, as those get forked automatically.

I can go ahead and add the cs10 chroots now.

nikic: The cs10 chroots are now present: https://copr.fedorainfracloud.org/coprs/g/fedora-llvm-team/llvm-test-suite/monitor/

jchecahi (approved):

tests/snapshot-gating.fmf
kwk: > The cs10 chroots are now present: https://copr.fedorainfracloud.org/coprs/g/fedora-llvm-team/llvm-test-suite/monitor/

@nikic awesome. Thank you!

PR #754, opened 2024-09-27
Adjust all chroots with `--with=snapshot_build`

kwk:
In order to avoid Copr specific code like the following in the `llvm.spec` file we have to manually specify the `--with=snapshot_build` for each chroot.

```spec
%global copr_snapshot_build_detected %(/bin/bash -fc 'if [[ "%{?copr_username}/%{?copr_projectname}" == @fedora-llvm-team/llvm-snapshots-* ]];
then
  echo "yes"
else
  echo "no"
fi')

%if "%{copr_snapshot_build_detected}" == "yes"
%bcond_without snapshot_build
%else
%bcond_with snapshot_build
%endif
```

This first began here: https://src.fedoraproject.org/rpms/llvm/pull-request/301#comment-221782

### DIFF
--- a/.github/workflows/fedora-copr-build.yml
+++ b/.github/workflows/fedora-copr-build.yml
@@ -130,13 +130,14 @@ jobs:
             --module-hotfixes on \
             $chroot_opts "${{ env.project_today }}"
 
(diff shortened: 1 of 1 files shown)


nikic (approved):
LGTM

PR #864, opened 2024-11-15
Add some automation for doing periodic mass rebuilds of Fedora packages

tstellar:
These mass rebuilds will help test the snapshot builds and catch regressions in real-world projects.  The GitHub workflows are currently configured to start a new mass rebuild once per month and then when the rebuild is complete, it will create any issue reporting any regressions since the last build.

The goal is to keep the rebuilder as simple as possible, so it will only rebuild packages that successfully built in the last rebuild, and it makes no attempt to work around packages that hard code gcc as their compiler.  As a result, some of the 'passing' builds may actually use gcc instead of clang, but this is an acceptable trade-off to keep the rebuild process simple.


kwk (changes requested):
I have some comments and probably need to look at this PR again to fully grasp it.

kwk (commented):
@tstellar I know that the CI system is buggy at the moment but I encourage you to add tests to your script. At least some doc tests which should be carried out as regular tests in our setup. I know this is hard for functions that involve copr and alike.

.github/workflows/mass-rebuild-reporter.yml
```diff
@@ -0,0 +1,80 @@
...
+            regressions.forEach(function(value){
+              comment = comment.concat('\n', value.name);
+              comment = comment.concat(': ', value.url);
+            });
+            console.log(comment);
+            const issue = await github.rest.issues.create({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               title: "Mass Rebuild Report",
+               labels: ['mass-rebuild'],
```
kwk, line 77: Please create `mass-rebuild` label first.

tstellar: Done: https://github.com/fedora-llvm-team/llvm-snapshots/labels/mass-rebuild

.github/workflows/rebuilder.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+def get_monthly_rebuild_packages(pkgs: set[str], copr_pkgs: list[dict]) -> set[str]:
+    for p in copr_pkgs:
+        latest_succeeded = p["builds"]["latest_succeeded"]
+        latest = p["builds"]["latest"]
+        if p["name"] not in pkgs:
+            continue
+        if not latest_succeeded:
+            pkgs.discard(p["name"])
```
tuliom, line 110: I wonder if we should have a small list of important packages that should never get removed even if they failed last time, e.g. qemu.

tstellar: I was planning to add tmt tests for important packages, like qemu, so that they get tested every day, not just periodically with the mass rebuild, but I agree, I think it does make sense to ensure they always get tested by this script too.

.github/workflows/rebuilder.py
```diff
@@ -0,0 +1,335 @@
...
+    >>> project_owner = "@fedora-llvm-team"
+    >>> project_name = "fedora41-clang-20"
+    >>> regressions = get_monthly_rebuild_regressions(project_owner, project_name, datetime.datetime.fromisoformat("2024-11-11"), copr_pkgs)
+    >>> print(regressions)
+    [{'name': 'f', 'url': 'https://copr.fedorainfracloud.org/coprs/@fedora-llvm-team/fedora41-clang-20/build/2/'}]
+
+    """
+    pkgs = []
+    for p in copr_pkgs:
+        latest_succeeded = p["builds"]["latest_succeeded"]
```
kwk, line 180: Please model this in some class.

.github/workflows/rebuilder.py
```diff
@@ -0,0 +1,335 @@
...
+    for p in pkgs:
+        logging.info("Rebuild", p)
+        copr_client.build_proxy.create_from_distgit(
+            project_owner, project_name, p, "f41", buildopts=buildopts
+        )
+
+
+def select_snapshot_project(
+    copr_client: copr.v3.Client, target_chroots: list[str], max_lookback_days: int = 14
+) -> str:
```
kwk, line 248:
```suggestion
def select_snapshot_project(
    copr_client: copr.v3.Client, target_chroots: list[str], max_lookback_days: int = 14
) -> str | None:
```

The function may return `None`.

README.adoc
```diff
@@ -196,3 +196,33 @@ $ make snapshot-rpm
 You might need to install missing dependencies. The build process itself probably takes quite some time.
 
 You're going to find `results/YYYYMMDD/snapshot-rpm.log` with logging everything from this makefile target.
+
+== Mass Rebuilds ==
+
+This repository uses GitHub Actions to periodically perform rebuilds of select
```
kwk, line 202:
```suggestion
This repository uses GitHub Actions to periodically perform rebuilds of selected
```

I'm not sure this is a typo because I'm no native speaker.

README.adoc
```diff
@@ -196,3 +196,33 @@ $ make snapshot-rpm
...
+Fedora packages.  Once a mass rebuild is complete there is also automation
+that will create a new issue with the results of the rebuild.  Note that this
+automation will run automatically when the issue for the previous month's mass
+rebuild is closed.  So don't close the previous months issue until the next
+months rebuild is complete.
```
kwk, line 207:
```suggestion
that will create a new issue with the results of the rebuild.

[NOTE] 
====
Note that this automation will run automatically when the issue
for the previous month's mass rebuild is closed. So don't close
the previous months issue until the next months rebuild is complete.
===
```

Let's use [admonitions](https://docs.asciidoctor.org/asciidoc/latest/blocks/admonitions/).

README.adoc
```diff
@@ -196,3 +196,33 @@ $ make snapshot-rpm
 You might need to install missing dependencies. The build process itself probably takes quite some time.
```
kwk, line 196: Let's enable [admonition icons](https://docs.asciidoctor.org/asciidoc/latest/blocks/admonitions/#enable-admonition-icons).

```suggestion
= README
:icons: font

You might need to install missing dependencies. The build process itself probably takes quite some time.
```

README.adoc
```diff
@@ -196,3 +196,33 @@ $ make snapshot-rpm
...
+that will create a new issue with the results of the rebuild.  Note that this
+automation will run automatically when the issue for the previous month's mass
+rebuild is closed.  So don't close the previous months issue until the next
+months rebuild is complete.
+
+The rebuild process will attempt to automatically bisect the failures to a specific upstream
+LLVM commit.
+
+The rebuild can be started manually using the rebuilder.py script in
+.github/workflows/
```
kwk, line 213:
```suggestion
The rebuild can be started manually using the rebuilder.py script in
`.github/workflows/`
```

README.adoc
```diff
@@ -196,3 +196,33 @@ $ make snapshot-rpm
...
+
+---
+$ python3 rebuilder.py rebuild
+---
```
kwk, line 217:
```suggestion
[source,console]
---
$ python3 rebuilder.py rebuild
---
```

README.adoc
```diff
@@ -196,3 +196,33 @@ $ make snapshot-rpm
...
+
+You can also view the regression report once the rebuild is complete using
+the same script.
+
+---
+$ python3 rebuilder.py get-regressions --start-date=<yyyy-mm-dd>
+---
```
kwk, line 224:
```suggestion
[source,console]
---
$ python3 rebuilder.py get-regressions --start-date=<yyyy-mm-dd>
---
```

.github/workflows/mass-rebuild-bisect.yml
```diff
@@ -0,0 +1,38 @@
+name: Mass Rebuild Bisect
+
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
+    inputs:
+      pkg:
+        description: "The package to bisect"
```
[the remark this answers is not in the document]
tstellar: Fedora package, I"ll update the description.

.github/workflows/rebuilder.py
```diff
@@ -0,0 +1,335 @@
...
+
+def get_pkgs(exclusions: set[str]) -> set[set]:
+    base = dnf.Base()
+    conf = base.conf
+    for c in "AppStream", "BaseOS", "CRB", "Extras":
+        base.repos.add_new_repo(
+            f"{c}-source",
+            conf,
+            baseurl=[
+                f"https://odcs.fedoraproject.org/composes/production/latest-Fedora-ELN/compose/{c}/source/tree/"
```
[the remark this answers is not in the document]
tstellar: Building all of Fedora would be too many packages, so we build Fedora ELN instead which is limited to only the packages that are part of RHEL.

.github/actions/prepare-python/action.yml
```diff
@@ -7,11 +7,16 @@ inputs:
     description: "Where this project was checked out"
     required: false
     default: "."
+  use-system-python:
```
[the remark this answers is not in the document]
tstellar: I was concerned I might break the other jobs that use the prepare-python action if I defaulted to system python.  Do you think they'll be OK?

.github/workflows/rebuilder.py
```diff
@@ -0,0 +1,335 @@
...
+        project_name,
+        with_latest_succeeded_build=True,
+        with_latest_build=True,
+    )
+
+
+def get_monthly_rebuild_packages(pkgs: set[str], copr_pkgs: list[dict]) -> set[str]:
+    """Returns the list of packages that should be built in the next rebuild.
+        It will select all the packages that built successfully during the last
+        rebuild.
```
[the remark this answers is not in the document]
tstellar: The pre-commit job is passing now.

README.adoc
tstellar: I think it's grammatically correct, but I'll adopt your suggestion since it is more readable.

.github/workflows/rebuilder.py
```diff
@@ -3,13 +3,66 @@
...
+
+    @property
+    def submitted_on(self):
+        return self.data["submitted_on"]
+
+
+class CoprPkg:
+
+    def __init__(self, data: dict):
+        self.data = data
```
kwk, line 35: I'm sorry for nitpicking but it is really hard to understand what needs to go into `data`. From looking in the code an element from the response of this [`copr_client.package_proxy.get_list`](https://python-copr.readthedocs.io/en/latest/client_v3/proxies.html#copr.v3.proxies.package.PackageProxy.get_list) method, needs to go here? The HTTP API response is documented [here](https://copr.fedorainfracloud.org/api_3/docs).

.github/workflows/rebuilder.py
```diff
@@ -91,16 +144,19 @@ def get_pkgs(exclusions: set[str]) -> set[set]:
...
+) -> list[CoprPkg]:
+    return [
+        CoprPkg(p)
+        for p in copr_client.package_proxy.get_list(
+            project_owner,
+            project_name,
+            with_latest_succeeded_build=True,
+            with_latest_build=True,
+        )
+    ]
```
kwk, line 156: Does is make sense to make this a class method of `CoprPkg`? At least topic wise they both belong together. 

Also, the function is currently called `get_builds_from_copr` but it returns a list of Copr packages. I suggest to rename it `get_packages_from_copr`. 

So something like:

```
@classmethod
def get_packages_from_copr(cls, project_owner: str, project_name: str, copr_client: copr.v3.Client) -> list["CoprPkg"]:
    return [
        CoprPkg(p)       
        for p in copr_client.package_proxy.get_list(
            project_owner,
            project_name,
            with_latest_succeeded_build=True,
            with_latest_build=True,
         )
    ]
```

.github/workflows/rebuilder.py
```diff
@@ -121,30 +177,28 @@ def get_monthly_rebuild_packages(pkgs: set[str], copr_pkgs: list[dict]) -> set[s
     >>> c = {"name" : "c", "builds" : { "latest" : { "id" : 2 } , "latest_succeeded" : { "id" : 1 } } }
     >>> d = {"name" : "d", "builds" : { "latest" : { "id" : 2 } , "latest_succeeded" : { "id" : 2 } } }
     >>> pkgs = { "b", "c", "d"}
-    >>> copr_pkgs = [a, b, c, d]
+    >>> copr_pkgs = [CoprPkg(p) for p in [a, b, c, d]]
```
kwk, line 180: I don't really like the more work of having to adjust test cased but it shows how awesome this doc tests really are. The documentation is code that MUST run.

.github/workflows/rebuilder.py
```diff
@@ -121,30 +177,28 @@ def get_monthly_rebuild_packages(pkgs: set[str], copr_pkgs: list[dict]) -> set[s
...
     for p in copr_pkgs:
-        latest_succeeded = p["builds"]["latest_succeeded"]
-        latest = p["builds"]["latest"]
-        if p["name"] not in pkgs:
+        if p.name not in pkgs:
             continue
-        if not latest_succeeded:
-            pkgs.discard(p["name"])
+        if not p.latest_succeeded:
+            pkgs.discard(p.name)
```
kwk, line 190: Now that I read this code I wonder if all you needed to do was to change the type hint of the `copr_pkgs` argument. Well, it is a hint so it wouldn't be necessary but I know that the [`Munch`](https://github.com/Infinidat/munch) type returned from the Copr API behaves like a `dict` but you can access elements like in objects.

In other words: The `p.name` access is possible even without having `CoprPkg`.

If you still want `CoprPkg` which I think makes sense you can experiment with:

```python
class CoprPkg(Munch):
    pass
```

or if you want to really show what properties there are:

```python
class CoprPkg(Munch):
    @property
    def name(self) -> str:
        return self.name
```

The same applies to `CoprBuild`. Note that you may need to wrap `Munch` responses in your own classes to fully work here. But that's worth it, depending on how much your classes actually can do.

.github/workflows/rebuilder.py
```diff
@@ -177,14 +231,11 @@ def get_monthly_rebuild_regressions(
...
-        latest_succeeded = p["builds"]["latest_succeeded"]
-        latest = p["builds"]["latest"]
-
-        if not latest:
+        if not p.latest:
             continue
 
         # Don't report regressions if there are still builds in progress
-        if latest["state"] not in [
+        if p.latest.state not in [
```
kwk, line 238: This test screems for a propery `def is_in_progress(self) -> bool`.

.github/workflows/rebuilder.py
```diff
@@ -329,6 +377,20 @@ def main():
             project_owner, project_name, start_time, copr_pkgs
         )
         print(json.dumps(pkg_failures))
+    elif args.command == "get-snapshot-date":
+        project = copr_client.project_proxy.get(project_owner, project_name)
+        for repo in project['additional_repos']:
+            match = re.match(r"copr://@fedora-llvm-team/llvm-snapshots-big-merge-([0-9]+)$", repo)
+            if not match:
+                continue
+            print(datetime.datetime.fromisoformat(match.group(1)).isoformat())
```
kwk, line 386: This prints more than one date so I suggest to rename it to `get-snapshot-dates`.

kwk (changes requested):
Hi @tstellar. Thank you for accepting so many of my suggestions. I've looked at your changes and I'm puzzled if it really was such a good idea to created `CoprBuild` and `CoprPkg` classes just to access properties like with objects. I think that before you had type hints that were simply wrong. It is a long time since I've last touched the Copr APIs but when I researched the responses of the Copr API I noticed that they don't return `dict`s but `Munch`s. And that allows for object like property access. For sure it helps to have less generic types like `Munch` but the mentioned subclassing in one of the comments is properly what I would go with.

scripts/rebuilder.py
```diff
@@ -381,14 +356,14 @@ def main():
...
-            print(datetime.datetime.fromisoformat(match.group(1)).isoformat())
+            if match:
+                print(datetime.datetime.fromisoformat(match.group(1)).isoformat())
+                return
     elif args.command == "rebuild-in-progress":
         for pkg in copr_client.monitor_proxy.monitor(project_owner, project_name)["packages"]:
             for c in pkg["chroots"]:
-                chroot = pkg["chroots"][c]
-                if chroot["state"] not in ["succeeded", "forked", "skipped", "failed", "canceled"]:
+                build = CoprBuild(pkg["chroots"][c])
```
kwk, line 365: The Copr API looks very weird, doesn't it? I mean you're iterating over chroots and then instantiate a `CoprBuild` variable from a chroot? Strange.

kwk:
@tstellar The only thing I ask for is that you cleanup the commit history a bit before merging, e.g. squashing the commits.

PR #933, opened 2024-12-15
Use correct edit-chroot option

nikic:
According to `copr edit-chroot --help`, the name of the option is `--repos`, not `--repo`.


nikic:
Hm, I think this is not it. In a local test just `--repo` does work and it does successfully modify the chroot. Not sure why this doesn't work in snapshot builds then...

nikic:
Ah, the reason that `--repo` works is that copr seems to generally accept prefixes of valid commands, to `--repos`, `--repo` and `--rep` etc all work. Doesn't explain what's going on with snapshots though.

PR #941, opened 2024-12-17
Consolitate matrix configuration

kwk:
I've created an action that lets us re-use a dynamic matrix configuration for both, building on copr and running checks.

The matrix from both touched workflows were simply merged and there is still some deduplication work ahead. But at least they share a common place now and one can ask for a matrix using workflow_dispatch events now.

Also, some of the low-hanging-fruit optimizations like `inputs.strategy == matrix.name` removal in `check-snapshots.yml` are not done to keep this PR reviewable.


kwk:
I could have waited for a review but I don't want to waste time and test this with tomorrow's build.

PR #944, opened 2024-12-18
Only run build and check workflows on org github repo

kwk:
We've identified people modifying the our org's copr repo by accident. This ensures for NEW forks, that this won't happen again by accident. Existing forks need to be adapted manually to avoid this modification.

### DIFF
--- a/.github/workflows/check-snapshots.yml
+++ b/.github/workflows/check-snapshots.yml
@@ -33,6 +33,7 @@ permissions:
 
 jobs:
   generate-matrix:
+    if: github.repository_owner == 'fedora-llvm-team'
     runs-on: ubuntu-latest
     outputs:
       mymatrix: ${{ steps.set-matrix.outputs.matrix }}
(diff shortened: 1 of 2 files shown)


nikic (changes requested):
I think it would be better to disable the workflows entirely using `if: github.repository_owner == 'fedora-llvm-team'`. I don't think people forking this repo actually want to run their own copy of the snapshots, right?

nikic:
> > I think it would be better to disable the workflows entirely using `if: github.repository_owner == 'fedora-llvm-team'`. I don't think people forking this repo actually want to run their own copy of the snapshots, right?
> 
> Agreed but I would still keep this change then.

I'm not really sure this makes sense. Note that this assumes that the github user name matches the copr name -- including the `@` prefix for groups. This is going to work for `@fedora-llvm-team`, but not for any of the user forks, even if the github and copr usernames happen to match.

kwk:
> > > I think it would be better to disable the workflows entirely using `if: github.repository_owner == 'fedora-llvm-team'`. I don't think people forking this repo actually want to run their own copy of the snapshots, right?
> > 
> > 
> > Agreed but I would still keep this change then.
> 
> I'm not really sure this makes sense. Note that this assumes that the github user name matches the copr name -- including the `@` prefix for groups. This is going to work for `@fedora-llvm-team`, but not for any of the user forks, even if the github and copr usernames happen to match.

Alright, alright you got me.

PR #960, opened 2024-12-23
Revert "Install python3-nanobind-devel without copr"

tuliom:
Temporarily revert commit be591b596d213d4020b0deef97e6385ba212fc6f while we wait for python-nanobind to reach stable in all supported Fedora versions.


tuliom:
I've just added an extra commit in order to avoid errors on Fedora 40.

PR #1020, opened 2025-01-16
mass-rebuild: Improve bisect functionality

tstellar:
The mass-rebuild-bisect workflow can now detect false positives and will update its status on the issue.  It will also only try to bisect when there is a regression on x86_64.

### DIFF
--- a/scripts/rebuilder.py
+++ b/scripts/rebuilder.py
@@ -237,10 +237,21 @@ def get_monthly_rebuild_regressions(
         pkgs.append(
             {
                 "name": p.name,
+                "fail_id" : p.latest.id,
                 "url": f"https://copr.fedorainfracloud.org/coprs/{owner_url}/{project_name}/build/{p.latest.id}/",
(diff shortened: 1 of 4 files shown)


scripts/rebuilder.py
```diff
@@ -369,6 +380,12 @@ def main():
         pkg_failures = get_monthly_rebuild_regressions(
             project_owner, project_name, start_time, copr_pkgs
         )
+        get_chroot_results(pkg_failures, copr_client)
+        # Delete attributes we don't need to print
+        for p in pkg_failures:
+            for k in ["fail_id", "chroots"]:
+                del p[k]
```
kwk, line 387: Why don't we need to print these?

kwk (approved):
LGTM.

scripts/rebuilder.py
tstellar: It's just extra information that we don't need, but I guess it wouldn't hurt to include them.

PR #1061, opened 2025-02-03
Set myself as the current maintainer

tuliom:


nikic:
@kwk Can you please disable the required checks on this repo? The python tests don't work for PRs from forks.

tuliom:
I've just created #1063 hoping to skip this issue.

kwk:
> @kwk Can you please disable the required checks on this repo? The python tests don't work for PRs from forks.

It basically has to do with how [protective](https://github.com/fedora-llvm-team/llvm-snapshots/settings/secrets/actions) github was with our secrets:

> They are not passed to workflows that are triggered by a pull request from a fork.

That said, please create branches for your PRs to the aforementioned repo in the main repo and not in your fork. For outside collaborators this isn't possible but for 99% of the work done in that repository it should be. You should all be able to push branches to the main repo, if not, please contact me.

I think we might get away without using secrets for our test runs if we used python's mock facility. I've documented a few examples in [this repository](https://github.com/kwk/python-mock-examples) because I wasn't sure I can get them to work in our repository. With these documented examples I think it is possible to mock github, copr and tmt with some effort. Then we wouldn't need any secrets and you could create PRs from your fork and run tests even locally.

But for now please create branches in the main repo.

PR #1063, opened 2025-02-04
Skip tests that require access to the token

tuliom:
Add function to GithubClient.has_token() and use it in order to skip tests that require access to a Github token.
This will allow to run a smaller set of tests from other forks and, most importantly, won't block their merge.


tuliom:
This PR is finally ready to be reviewed.

nikic (approved):
This looks good to me, but @kwk will probably want to take a look.

kwk (approved):
LGTM

PR #1066, opened 2025-02-06
Stop quoting multiple parameters

tuliom:
$chroot_opts holds multiple parameters that mustn't be treated as a single one.


tuliom:
shellcheck, word splitting is exactly what we need.

kwk (approved):
LGTM

kwk (changes requested):
Please add a shellcheck disable comment

kwk (approved):

PR #1067, opened 2025-02-07
Allow after_build_id to be empty

tuliom:
Stop quoting after_build_id in order to use it even when it's empty. Disable shellcheck error reporting on this.


nikic (commented):
Maybe we can drop this entirely, now that we're only building one package?

tuliom:
@nikic Makes sense. I've just incorporated that suggestion to this PR.

kwk:
Closed in favor of https://github.com/fedora-llvm-team/llvm-snapshots/pull/1069

PR #1069, opened 2025-02-10
Restructure to only build the llvm package and not more

kwk:
With the advent of big-merge we only need to build one package, namely llvm. This restructures the github workflows and the actions used in them to only deal with the llvm package. The python scripts were not touched. yet.

This supersedes #1067


tuliom:
@kwk Interesting.. I'm not able to merge PRs without reviews. Am I missing anything?

PR #1083, opened 2025-02-13
rebuilder: Always rebuild important packages

tstellar:
Add a list of 'tier0' package that should always be rebuilt, even if their previous build failed.


tstellar:
Suggested by @tuliom .

tuliom (approved):

PR #1091, opened 2025-02-18
rebuilder: Bump version of clang to 21

tstellar:


tuliom (approved):
LGTM.
I wonder if we should move to f42 too.

Regarding the failures in the coverage test: there appears to be something wrong with the test itself.

PR #1102, opened 2025-02-24
[workflows] reorg

kwk:
* The `python-format-and-tests.yml` workflow did too many jobs in one workflow and didn't allow us to treat each job it did with different care.

* We want code to be formatted with `black` and we want this to be mandatory in code reviews. Let's do this in `check-python-code-formatting-with-black.yml` from now on.

* We want python tests to pass, so let's do this in `check-python-code-with-pytest.yml`.

* We want to generate coverage but we potentially don't want to make this mandatory for pull requests. For example, see 
image
 for this PR which doesn't change anything regarding coverage numbers. See `check-python-code-coverage.yml`.

* The `retest.yml` just got a new name: `retest-chroots-on-testing-farm.yml`.

* The `Makefile` got dedicated target we can use from within the workflows.

In order to unblock this PR I first have to get approval for it and then I can tune the branch protection settings to no longer require these old workflows image but the new ones

image


tuliom:
There are 2 tests pending (`run-tests` and `check-python-with-black`).
Does this PR need to be merged before these tests are executed?

kwk:
> > We want code to be formatted with black and we want this to be mandatory in code reviews.
> 
> I don't remember ever discussing this part with the rest of the team. IMHO, it's a good idea to have a tool that helps to check code formatting, but I've seen this test fail with super-strict results, e.g. complaining about how the lines of imports should be organized/grouped together.

Actually, the pre-commit git hook which should run on every `git commit` locally will complain about unordered imports etc. and reformat it to match the expectation. I've introduced `black` probably years ago to not have discussions about any formatting whatsoever.

> Is it possible to reduce that requirement?

IMHO. No because I don't see why not to follow the idea that `black` has.

kwk:
> There are 2 tests pending (`run-tests` and `check-python-with-black`). Does this PR need to be merged before these tests are executed?

The two pending tests are the ones, that I've split up in this PR:

1. [check-python-with-black](https://github.com/fedora-llvm-team/llvm-snapshots/blob/3d27c77571f3144d9c3864e057fa848b9b2ae4b5/.github/workflows/python-format-and-tests.yml#L8-L15) is now [here](https://github.com/fedora-llvm-team/llvm-snapshots/pull/1102/files#diff-82cc24fa1c1234755c2862aecf32b0192e6190f88eaa79ac0a48ed1dc704d281R8-R15).
2. [run-tests](https://github.com/fedora-llvm-team/llvm-snapshots/blob/3d27c77571f3144d9c3864e057fa848b9b2ae4b5/.github/workflows/python-format-and-tests.yml#L16-L35) is now [here](https://github.com/fedora-llvm-team/llvm-snapshots/blob/3d27c77571f3144d9c3864e057fa848b9b2ae4b5/.github/workflows/python-format-and-tests.yml#L16-L35) and [here](https://github.com/fedora-llvm-team/llvm-snapshots/blob/3d27c77571f3144d9c3864e057fa848b9b2ae4b5/.github/workflows/python-format-and-tests.yml#L16-L35).

When this PR gets approved I'm going to adjust the branch protection rule to remove the two tests and instead add the pytest run and black formatting individually. At least that was the plan.

tuliom (approved):
While I disagree about the code style and I believe it should be relaxed, I have to recognize this PR is not changing the current behavior.
So, LGTM.

kwk:
The branch protection rules for `main` are now changed to this:

image

Unfortunately the old branch protection way of configuring things only shows the job names and not the workflow names. `black` and `pytest` are the correct ones to choose though.

In the future we might wanna look into [rulesets](https://github.com/fedora-llvm-team/llvm-snapshots/settings/rules).

Issue #1046, opened 2025-01-27
Remove obsolete build time diagrams


nikic:
The currently link to: https://fedora-llvm-team.github.io/llvm-snapshots/fig-llvm.html

I think we can drop all the stale data for standalone builds and also link to the big-merge figure by default.

PR #1107, opened 2025-02-24
Remove obsolete build time diagrams

kwk:
This change does different things at once.

1. No longer use `build-stats.csv` but only `build-stats-big-merge.csv` and `build-stats-pgo.csv`. The latter has been there before and might be useful in the future.

2. The `index.html` landing page with an overview of packages got removed because with the advent of big-merge we only have the llvm package.

3. The new `index.html` is what formerly was the `fig-big-merge.html`.

4. The `Makefile` contains a new recipe for you to produce the diagrams locally for testing.

5. The "big-merge" way of building allowed me to remove a lot of code which previously was needed to merge individual packages together.

This is how the new page looks: 
image

Fixes: #1046

### DIFF
--- a/scripts/create-diagrams.py
+++ b/scripts/create-diagrams.py
@@ -2,7 +2,6 @@
 
 # %%
 import argparse
-import re
 from datetime import datetime
 from pathlib import Path
 
@@ -11,7 +10,6 @@ import pandas as pd
 import plotly.express as px
 import plotly.graph_objects as go
 import plotly.io as pio
-from plotly.offline import plot
 
 
 # %%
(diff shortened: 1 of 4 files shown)


scripts/create-diagrams.py
```diff
@@ -115,47 +113,35 @@ def save_figure(fig: go.Figure, filepath: str) -> None:
...
+    header_menu += f" <small>(Last updated: {last_updated})</small>"
     header_menu += "</div>"
     header_menu += replace_me
 
     file.write_text(file.read_text().replace(replace_me, header_menu))
 
 
 # %%
-def prepare_data(filepath: str = "build-stats.csv") -> pd.DataFrame:
+def prepare_data(filepath: str = "build-stats-big-merge.csv") -> pd.DataFrame:
```
nikic, line 140: nit: Remove default value? You don't make use of it.

nikic (approved):
LGTM

scripts/create-diagrams.py
kwk: Done in 3c20f62.

PR #1108, opened 2025-02-24
[build-diagrams] show successful builds by default

kwk:
* I've removed the PGO builds from the menu because I'm not working on it atm. The files themselves are still written.

* I've added two options to view big-merge stats:

  1. *Only successful builds*, which is the default, will give a very good overview of build time evolution.

image

  3. *All states* shows builds that ended in all possible states (that's what we had before).

image

### DIFF
--- a/scripts/create-diagrams.py
+++ b/scripts/create-diagrams.py
@@ -127,8 +127,12 @@ def add_html_header_menu(filepath: str, plotly_div_id: str = "plotly_div_id") ->
 
     file = Path(filepath)
     header_menu = '<div id="headermenu">Build-Stats: '
-    header_menu += ' <a href="index.html">llvm (big-merge)</a>'
-    header_menu += ' | <a href="fig-pgo.html">llvm (pgo)</a>'
+    header_menu += ' <a href="index.html">big-merge (only success)</a>'
+    header_menu += (
+        ' | <a href="fig-big-merge-all-states.html">big-merge (all states)</a>'
+    )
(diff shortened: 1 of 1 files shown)


scripts/create-diagrams.py
```diff
@@ -205,29 +209,47 @@ def main() -> None:
         "plotly"  # See https://plotly.com/python/templates/#theming-and-templates
     )
 
-    # Create dataframe of llvm in "big-merge mode". The chroots are prefixed
-    # with "big-merge-" on the fly to be able to distinguish the two cases.
+    # # Create dataframe of llvm in "big-merge mode". The chroots are prefixed
+    # # with "big-merge-" on the fly to be able to distinguish the two cases.
```
nikic, line 213: Why the double comments here?

nikic (dismissed):

scripts/create-diagrams.py
```diff
@@ -205,29 +209,47 @@ def main() -> None:
...
     df_big_merge = prepare_data(filepath=args.datafile_big_merge)
     df_big_merge["chroot"] = "big-merge-" + df_big_merge["chroot"]
-    # Convert build_id column with int64's in it to an array of int64's.
+    # # Convert build_id column with int64's in it to an array of int64's.
```
tuliom, line 216: Likewise here.

tuliom (commented):
LGTM. I believe there is a typo, though.

PR #1114, opened 2025-02-26
Limit the number of fedora-s390x chroots

tuliom:
COPR restricts the number of s390x builds each team can start, queueing builds after this limit is reached. This is causing daily builds to take too long to complete.

In order to minimize this, we decided to limit the number of fedora-s390x chroots to 2: rawhide and the latest branched version.

Add an exclude mechanism to get_chroots() that allows to apply this limit.

### DIFF
--- a/scripts/functions.sh
+++ b/scripts/functions.sh
@@ -54,7 +54,18 @@ function get_active_build_ids(){
 
 # Prints the chroots we care about.
 function get_chroots() {
-  copr list-chroots | grep -P '^(fedora-(rawhide|[0-9]+)|rhel-[8,9]-)' | sort |  tr '\n' ' '
+  local tmp=`mktemp`
+  copr list-chroots | grep -P '^(fedora-(rawhide|[0-9]+)|rhel-[8,9,10]+-)' > $tmp
+  # We can run at most 4 s390x builds at the same time on COPR.
+  # In order to avoid queueing too many s390x builds, we decided to test at
+  # most 2 fedora-s390x (rawhide and the latest branched version).
+  local excluded=`grep '^fedora-.*-s390x' $tmp | sort -urV | tail --lines=+3`
+  local excluded_regex=''
+  for e in $excluded; do
+    excluded_regex="$excluded_regex${excluded_regex:+|}$e"
+  done
+  grep -vE "$excluded_regex" $tmp | sort |  tr '\n' ' '
+  rm $tmp
 }
 
 # Returns 0 if all llvm packages on all chroots have successful builds.


kwk:
@tuliom for building snapshots, it is enough to change `get_chroot` but unfortunately this is not the only place we have to adjust (I wish it was though). We have another chroot pattern setting [here](https://github.com/fedora-llvm-team/llvm-snapshots/blob/7b854e4038e56dd332f1fd1a1001d80dc6ae4217/.github/actions/generate-matrix/action.yml#L35C15-L35C73):

```yaml
"chroot_pattern": "^(fedora-(rawhide|[0-9]+)|rhel-[8,9]-)"
```

 In the [`check-snapshots.yml`](https://github.com/fedora-llvm-team/llvm-snapshots/blob/7b854e4038e56dd332f1fd1a1001d80dc6ae4217/.github/workflows/check-snapshots.yml#L86-L88) we pass the `chroot-pattern` option

```bash
          if [[ -n "${{ matrix.chroot_pattern }}" ]]; then
            extra_args="${extra_args} --chroot-pattern ${{matrix.chroot_pattern}}"
          fi
```


 to the [`snapshot_manager/main.py`](https://github.com/fedora-llvm-team/llvm-snapshots/blob/7b854e4038e56dd332f1fd1a1001d80dc6ae4217/snapshot_manager/main.py#L108-L115) program:

```python
    subparser_check.add_argument(
        "--chroot-pattern",
        metavar="REGULAR_EXPRESSION",
        type=str,
        dest="chroot_pattern",
        default=cfg.chroot_pattern,
        help="Chroots regex pattern for required chroots.",
    )
```

 Internally, [this function](https://github.com/fedora-llvm-team/llvm-snapshots/blob/7b854e4038e56dd332f1fd1a1001d80dc6ae4217/snapshot_manager/snapshot_manager/copr_util.py#L138-L160) is responsible for producing a list of chroot names that we check for:

```python
    def get_copr_chroots(self, pattern: str | None = None) -> list[str]:
        """Return sorted list of chroots we care about

        If no pattern is supplied the config variable `chroot_pattern` will be used.

        Args:
            pattern (str|None, optional): Regular expression to filter chroot names by. Defaults to None.

        Returns:
            list[str]: List of filtered and sorted chroots
        """
        if pattern is None:
            pattern = self.config.chroot_pattern

        if self._all_chroots is None:
            self._all_chroots = self.copr.mock_chroot_proxy.get_list().keys()

        chroots = []
        for chroot in self._all_chroots:
            if re.match(pattern=pattern, string=chroot) != None:
                chroots.append(chroot)
        chroots.sort()
        return chroots
```

I know that this duplication is quite unfortunate. But in my defense so far I was able to copy the regular expression to two places and had no problems.

Before my sick leave I was working on a refactoring of the whole copr functionality in the snapshot-manager. The functions were not all used and overall things could have been made simpler. In the new code the messy `get_copr_chroots` function from above [looks cleaner](https://github.com/kwk/llvm-snapshots/blob/copr-refactor/snapshot_manager/snapshot_manager/copr_util.py#L42-L78) and is split in two:

```python
@functools.cache
def get_all_chroots(client: copr.v3.Client) -> list[str]:
    """Asks Copr to list all currently supported chroots. The response Copr will
    give varies over time whenever a new Fedora or RHEL version for example
    is released. But for our purposes, we let the function cache the results.

    Args:
        client (copr.v3.Client): A Copr client

    Returns:
        list[str]: All currently supported chroots on copr.
    """
    return client.mock_chroot_proxy.get_list().keys()


def filter_chroots(chroots: list[str], pattern: str) -> list[str]:
    """Return a sorted list of chroots filtered by the given pattern.

    Args:
        chroots (list[str]): As list of chroots to filter
        pattern (str, optional): Regular expression e.g. `r"^(fedora-(rawhide|[0-9]+)|rhel-[8,9]-)"`

    Returns:
        list[str]: List of filtered and sorted chroots.

    Examples:

    >>> chroots = ["rhel-7-x86_64", "rhel-9-s390x", "fedora-rawhide-x86_64", "centos-stream-10-ppc64le"]
    >>> filter_chroots(chroots=chroots, pattern=r"^(fedora-(rawhide|[0-9]+)|rhel-[8,9]-)")
    ['fedora-rawhide-x86_64', 'rhel-9-s390x']
    """
    res: list[str] = []
    for chroot in chroots:
        if re.match(pattern=pattern, string=chroot) != None:
            res.append(chroot)
    res.sort()
    return res
```

Technically it would be ideal if we can move all the logic that you've written in bash in python and only have one place to change this.

kwk:
I'm rewriting the [`generate-matrix.yml`](https://github.com/fedora-llvm-team/llvm-snapshots/blob/7b854e4038e56dd332f1fd1a1001d80dc6ae4217/.github/actions/generate-matrix/action.yml#L22-L72) part that is in bash in python now. Then we at least can share more code.

kwk:
For starters, here's some sanitization code that we can use inside python.

```python

from snapshot_manager.util import chroot_version, expect_chroot

def sanitize_chroots(chroots:list[str]) -> list[str]:
    """Removes all s390x fedora chroots but rawhide and the latest numbered version"""
    chroots = [expect_chroot(chroot) for chroot in chroots]
    removals=filter_chroots(chroots, r"^fedora-[0-9]+-s390x")
    max = 0
    for chroot in removals:
        ver = int(chroot_version(chroot))
        if ver > max:
            max = ver
    for chroot in removals:
        if max != int(chroot_version(chroot)):
            chroots.remove(chroot)
    return chroots
```

kwk:
Once https://github.com/fedora-llvm-team/llvm-snapshots/pull/1146 lands, and we have the `sanitize_chroots` function, then this PR can be closed.

PR #1162, opened 2025-03-12
Attempt to mitigate lsb_release error

kwk:
We saw this error before:

```
Error: Unable to locate executable file: lsb_release. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.
```

image


kwk:
Closing this because the real issue that the prepare-python action is now called from a fedora:41 container. I'm going to replace this there.

kwk:
See #1163 instead.

PR #1163, opened 2025-03-12
Attempt to mitigate lsb_release error

kwk:
We saw this error before:

```
Error: Unable to locate executable file: lsb_release. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.
```

The root cause is that the `prepare-python` action is tailored to run on `ubuntu` containers. As a short term fix for this I've implemented the steps manually in the `fedora-copr-build.yml` workflow which runs on a `fedora:41` container.

### DIFF
--- a/.github/workflows/fedora-copr-build.yml
+++ b/.github/workflows/fedora-copr-build.yml
@@ -54,8 +54,28 @@ jobs:
     container: fedora:41
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


tstellar (approved):

kwk:
Unfortunately this PR didn't solve the problem:

image

PR #1164, opened 2025-03-12
Prevent empty strategy in scheduled workflow run

kwk:
When the the workflow `fedora-copr-build.yml` runs on a schedule the `inputs` are empty and no strategy will be provided. In this case we want to run all strategies.

This should fix this error: `main.py github-matrix: error: argument --strategy: expected one argument`

image

For now we don't want pgo builds so I've commented out the config for them.

### DIFF
--- a/snapshot_manager/main.py
+++ b/snapshot_manager/main.py
@@ -56,7 +56,6 @@ def main():
     if cmd == "check":
         cfg.github_repo = args.github_repo
         cfg.datetime = args.datetime
-        cfg.strategy = args.strategy
         snapshot_manager.SnapshotManager(config=cfg).check_todays_builds()
     elif cmd == "retest":
         cfg.github_repo = args.github_repo
@@ -135,7 +134,7 @@ def add_strategy_argument(argparser: argparse.ArgumentParser) -> argparse.Action
         "--strategy",
         dest="strategy",
         type=str,
-        default="",
+        required=True,
         help=f"Strategy to use",
     )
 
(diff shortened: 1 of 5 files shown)


kwk:
> https://github.com/fedora-llvm-team/llvm-snapshots/blob/ef2d28fda4c8ae2b705f577a3268a4812e6a027b/snapshot_manager/snapshot_manager/util.py#L671
> handles both `"all"` and `""`. Can we normalize on one of them to mean "all strategies"?

Sure. I've created a new issue for this.

nikic:
It's directly related to what you are doing in this PR, not a separate task.

kwk:
> It's directly related to what you are doing in this PR, not a separate task.

Okay, I've addressed it here in [116ccba](https://github.com/fedora-llvm-team/llvm-snapshots/pull/1164/commits/116ccbadac5125e555c38cccdafdfed57bdb2804)

snapshot_manager/main.py
```diff
@@ -56,7 +56,6 @@ def main():
     if cmd == "check":
         cfg.github_repo = args.github_repo
         cfg.datetime = args.datetime
-        cfg.strategy = args.strategy
```
kwk, line 59: There is no `strategy` member in `config.Config` but only `build_strategy`. The strategy is already filled when selecting the strategy [here](https://github.com/fedora-llvm-team/llvm-snapshots/pull/1164/commits/116ccbadac5125e555c38cccdafdfed57bdb2804#diff-e735df349e8a75dd96a90e260766907bedb904e7828748af2981ef729e027288L48-L54).

nikic (approved):

PR #1235, opened 2025-04-04
bisect.sh: Download pre-built clang from COPR to check for false positives

tstellar:
The COPR builds will not always be available, but when they are, this will save us 1 or 2 llvm rebuilds when trying to bisect a failure.

### DIFF
--- a/scripts/bisect.sh
+++ b/scripts/bisect.sh
@@ -9,6 +9,71 @@ function get_clang_commit {
   curl "https://download.copr.fedorainfracloud.org/results/@fedora-llvm-team/fedora-41-clang-21/fedora-41-x86_64/0$buildid-$pkg/root.log.gz" | gunzip |  grep -o 'clang[[:space:]]\+x86_64[[:space:]]\+[0-9a-g~pre.]\+' | cut -d 'g' -f 3
 }
 
+function get_clang_copr_project {
+  buildid=$1
+  pkg=$2
+  arch=$(rpm --eval %{_arch})
+
+  date=$(curl "https://download.copr.fedorainfracloud.org/results/@fedora-llvm-team/fedora-41-clang-21/fedora-41-$arch/0$buildid-$pkg/root.log.gz" | gunzip |  grep -o "clang[[:space:]]\+$arch[[:space:]]\+[0-9.]\+~pre[0-9]\+" | cut -d '~' -f 2 | sed 's/pre//g')
+  echo "@fedora-llvm-team/llvm-snapshots-big-merge-$date"
+}
+
+function configure_llvm {
+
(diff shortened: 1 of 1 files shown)


scripts/bisect.sh
```diff
@@ -34,27 +99,31 @@ srpm_url=$(curl -X 'GET' "https://copr.fedorainfracloud.org/api_3/build/$buildid
...
-# Test the good commit to see if this a false positive
-git checkout $good_commit
+# Enable the compat libraries so we can install a clang snapshot.
+# We need to do this because the runtime repo dependencies from
+# the snapshot copr projects is not resolved when using dnf5 see:
+# https://github.com/fedora-copr/copr/issues/3387
+dnf copr enable -y @fedora-llvm-team/llvm-compat-packages
 
-cmake -G Ninja -B build -S llvm -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_PROJECTS=clang -DLLVM_TARGETS_TO_BUILD=Native -DLLVM_BINUTILS_INCDIR=/usr/include/ -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER=/opt/llvm/bin/clang++ -DCMAKE_C_COMPILER=/opt/llvm/bin/clang
+# Test the good commit to see if this a false positive.
```
tuliom, line 109: There is a typo in s/this a false/this is a false/
But it took me a while to understand what you meant with "false positive" here.
I wonder if the following is clearer:
```
# Test if the good commit still succeeds. A failure may indicate an
# intermittent failure or an issue that is unrelated to LLVM. In either case,
# this is not a "good commit".
```

scripts/bisect.sh
```diff
@@ -9,6 +9,71 @@ function get_clang_commit {
...
+
+if [ -n "$LLVM_SYSROOT" ] -a [ -e "$LLVM_SYSROOT/bin/clang" ]; then
+  cc=$LLVM_SYSROOT/bin/clang
+  cxx=$LLVM_SYSROOT/bin/clang++
+else
+  cc=clang
+  cxx=clang++
+fi
+
+cmake -G Ninja -B build -S llvm -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_PROJECTS=clang -DLLVM_TARGETS_TO_BUILD=Native -DLLVM_BINUTILS_INCDIR=/usr/include/ -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER=$cxx -DCMAKE_C_COMPILER=$cc
```
tuliom, line 31: All lines in this function are missing the first indentation level.

tuliom (changes requested):
The logic LGTM.
I found 2 cosmetic issues, though.

tuliom (approved):

PR #1249, opened 2025-04-10
Build the reproducer container on all arches

tstellar:


Containerfile.mass-rebuild
```diff
@@ -1,6 +1,12 @@
-FROM registry.fedoraproject.org/fedora:41
-
-ENV LLVM_SYSROOT=/opt/llvm \
+ARG arch=x86_64
```
tuliom, line 1: I don't understand what will happen when building for other architectures.
Is this assignment ignored?

.github/workflows/build-reproducer-container.yml
```diff
@@ -20,8 +27,22 @@ jobs:
...
+          if [ -n "$llvm_container_arch" ]; then
+            extra_args="--build-arg optimized_llvm=ghcr.io/llvm/$llvm_container_arch/ci-ubuntu-22.04:9f919661dd2f --build-arg llvm_sysroot=/opt/llvm/"
+          fi
+
           sudo apt-get -y update
-          sudo apt-get -y install podman
-          podman build -t "ghcr.io/$GITHUB_REPOSITORY_OWNER/llvm-snapshots-reproducer" -f Containerfile.mass-rebuild scripts/
+          sudo apt-get -y install podman qemu-user-static
+          # shellcheck disable=SC2086
+          podman build  --arch ${{matrix.arch }} --build-arg arch=${{ matrix.arch }} $extra_args -t "ghcr.io/$GITHUB_REPOSITORY_OWNER/llvm-snapshots-reproducer:${{ matrix.arch }}" -f Containerfile.mass-rebuild scripts/
```
tuliom, line 46: I learned to use `--platform` instead of `--arch` because it also sets other information, including `os` and `variant`.
I've already seen some weird cases where the environment in the container got mixed up.

Containerfile.mass-rebuild
```diff
@@ -1,6 +1,12 @@
...
+ARG base_image=registry.fedoraproject.org/fedora:41-${arch}
```
tuliom, line 2: AFAIU, if `--platform` is used, you don't need to specify the architecture here.

tuliom (commented):
The code LGTM in overall.
I have a question and a suggestion that might help to simplify it a little bit.

Containerfile.mass-rebuild
tstellar: That's just the default for when someone doesn't specify they arch option, but I've dropped this now based on your suggestion below.

.github/workflows/build-reproducer-container.yml
tstellar: Thanks for the tip about `--platform`` that worked well and simplified the containerfile.

tuliom (approved):

PR #1252, opened 2025-04-10
Create perf-results.csv with initial data

kwk:
We're going to append to this file later with automation in github workflows. For now, adding this file does not hurt because it is not used anywhere.

This is a PR agains the `gh-pages` branch which we use for storing the our build time HTML files and CSV files. This branch is updated nightly through automation when gathering the latest build information.

> [!NOTE]  
> Github renders CSV files as a table. That means. you get the best review experience when you take a look at the CSV file by directly heading over to the `add-initial-perf-results.csv` branch and looking at the file there: https://github.com/fedora-llvm-team/llvm-snapshots/blob/add-initial-perf-results.csv/perf-results.csv

Some of the information in this CSV file will later be used to render little hover boxes for each datapoint in a performance diagram. Here's an example of what I'm talking about:

image

We capture more CPU info than we actually display simply because I cannot tell which effect i.e. vulnerability mitigations can have on performance. If we later see spikes or drops in performance, having these CPU information readily available might help to draw some conclusions.


tuliom (approved):
I see this is going to be stored in branch `gh-pages`.
I assume the robot will push directly to this branch without human intervention.
This may increase the size of the repository over time, but I trust that compression will be efficient.
LGTM.

kwk:
> I see this is going to be stored in branch `gh-pages`. I assume the robot will push directly to this branch without human intervention. This may increase the size of the repository over time, but I trust that compression will be efficient. LGTM.

You're right the "robot"  will push to this branch without human intervention. That's what we currently do for build time diagrams. If we feel that some information in the CSV file are not needed we can always remove columns that consume a lot of space.

PR #1269, opened 2025-04-10
[workflow] Add start-performance-comparison.yml

kwk:
* #1282
* #1281
* #1280
* #1279
* #1278
* #1277
* #1276
* #1275
* #1274
* #1273
* #1272
* #1271
* #1270
* __->__ #1269
* #1268

We intend to run performance comparisons on a daily basis. This new
workflow does the first step by running every night close after
midnight.

As a simplification we assume that after 24 hours we have all the
`pgo` and `big-merge` COPR builds ready from the day before.

The new `start-perf-comparison` of the the `main.py` program will be
added in a later PR. To make it easy to follow the PRs I decided to
bring in the necessary changes piece by piece.

### DIFF
--- a/.github/workflows/start-performance-comparison.yml
+++ b/.github/workflows/start-performance-comparison.yml
@@ -1,0 +1,26 @@
+name: "Start performance comparison"
+
+on:
+  schedule:
+    # Everyday at 00:45am
+    # See https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule
+    - cron: "45 0 * * *"
+
+  workflow_dispatch:
+
+jobs:
+  start-performance-comparison:
+    if: github.repository_owner == 'fedora-llvm-team'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/prepare-python
+      - name: "start-performance-comparison"
+        run: |
+          today=$(date +%Y%m%d)
+          yesterday=$(date -d "${today} -1 day" +%Y%m%d)
+
+          python3 ./snapshot_manager/main.py start-perf-comparison \
+            --strategy-a pgo \
+            --strategy-b big-merge \
+            --yyyymmdd "$yesterday"


.github/workflows/start-performance-comparison.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/prepare-python
+      - name: "start-performance-comparison"
+        run: |
+          today=$(date +%Y%m%d)
+          yesterday=$(date -d "${today} -1 day" +%Y%m%d)
+
+          python3 ./snapshot_manager/main.py start-perf-comparison \
```
tuliom, line 23: This command is not available in the repository yet.
Is this commit out-of-order?

tuliom (changes requested):

.github/workflows/start-performance-comparison.yml
kwk: The PR body says that it will be added later:

> The new `start-perf-comparison` of the the `main.py` program will be
> added in a later PR. To make it easy to follow the PRs I decided to
> bring in the necessary changes piece by piece.

tuliom: Right, but isn't this is adding broken code?
Notice that I'm not suggesting to merge this commit with another, but I expected it should come after the code it depends on is implemented.
Does that make sense?

tuliom (approved):
My previous assumptions might be incorrect if this PR is merged as a squashed commit with the rest of the stacked PRs.
Even if it isn't squashed, the impact to developers will be limited because this is inside a Github workflow.
With that said, I'm approving this.

.github/workflows/start-performance-comparison.yml
kwk: @tuliom this does make sense yes. In software development we technically can only develop bottom up (unless we mock the lower parts), which is what I've done while developing. But I thought that for reviewing code, this is not necessarily the easiest way to structure reviews and guide the reader along. I did refactor code while integrating new features that demanded the refactoring in the first place. And tearing this apart in the aftermath is quite complicated. In the end I squashed everything into one commit and soft reset to `HEAD^`, Then I added the things piece by piece as I saw fit. If you want me in the future to produce self-contained PRs, I can do my best. For now I figured the way I did it is the best I can do. I do understand that it is hard to do git bisect with a bunch of stacked PRs but should a bisect bring up one of the stacked PRs commits, you should be able to draw a conclusion that you should skip the whole stack and look before and after. I know, this means more work. What can I say other than that I'm sorry for how my brain works ;)

PR #1271, opened 2025-04-10
Add run-perf-comparison, collect-perf-results cmds

kwk:
* #1282
* #1281
* #1280
* #1279
* #1278
* #1277
* #1276
* #1275
* #1274
* #1273
* #1272
* __->__ #1271
* #1270
* #1269
* #1268

This adds the `run-perf-comparison` and `collect-perf-results`
entrypoints to the `main.py` program.

### DIFF
--- a/snapshot_manager/main.py
+++ b/snapshot_manager/main.py
@@ -106,7 +106,7 @@ def main():
             logging.warning("Not all builds were successful")
             sys.exit(1)
         logging.info("All required builds were successful")
-    elif cmd == "run-perf-comparison":
+    elif cmd in ("run-perf-comparison", "collect-perf-results"):
         github_token = os.getenv(config.Config().github_token_env)
         if github_token is None or len(github_token) == 0:
             logging.error(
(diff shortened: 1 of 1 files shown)


tuliom (changes requested):
This PR depends on `run_performance_comparison` without providing it. That is causing 4 Github checks to fail.
This is being added in PR #1272.
However, the same checks are also failing on that PR.
In order to avoid wasting time, I believe it's OK to merge this PR in what I call "inverted order".
But before that happens, could you check the test results from PR #1272, please? We need to understand why it's still failing because there is a chance this PR is adding a regression.

tuliom:
It looks like the issue is really in PR #1272, but it would be great to see it passing the checks (or failing due to dependency on another commit) before this is merged.

kwk:
> This PR depends on `run_performance_comparison` without providing it. That is causing 4 Github checks to fail. This is being added in PR #1272. However, the same checks are also failing on that PR. In order to avoid wasting time, I believe it's OK to merge this PR in what I call "inverted order". But before that happens, could you check the test results from PR #1272, please? We need to understand why it's still failing because there is a chance this PR is adding a regression.

With the inverted order, none of the tests probably work except for [top most PR](https://github.com/fedora-llvm-team/llvm-snapshots/pull/1282). The intermediate PRs fail because not only were the functions missing and then added in #1272 but the refactoring of the testing-farm stuff was missing as well. But honestly this should not hurt the review. A `tf.TestingFarmRequest` was changed to `tf.Request` for example because the `tf` stands for the `testing_farm` module, so there was no need to repeat this in the class name. Those are the cosmetic changes that I interleaved as I went. I promise to do better next time and have each PR in a stack pass tests. But then I would argue we don't need a stack.

tuliom (approved):

PR #1272, opened 2025-04-10
Add performance entrypoint fns for run + collect

kwk:
* #1282
* #1281
* #1280
* #1279
* #1278
* #1277
* #1276
* #1275
* #1274
* #1273
* __->__ #1272
* #1271
* #1270
* #1269
* #1268

This change adds these three functions

```
run_performance_comparison
collect_performance_comparison_results
get_performance_github_issue
```

### DIFF
--- a/snapshot_manager/snapshot_manager/snapshot_manager.py
+++ b/snapshot_manager/snapshot_manager/snapshot_manager.py
@@ -485,3 +485,256 @@ class SnapshotManager:
         )
         logging.info(f"Adding label: {labels_to_add}")
         issue.add_to_labels(*labels_to_add)
+
+
+def run_performance_comparison(
+    conf_a: config.Config,
+    conf_b: config.Config,
+    github_repo: str,
+    copr_client: copr.v3.Client,
+    github_client: github.Github,
+) -> bool:
+    """Runs a performance test for the two given configurations.
+
+    NOTE: We will only run tests for the those chroots that are shared by both configurations.
+
+    A github issue will be created that contains links and hidden HTML comments
+    testing-farm request info for all chroot configs that were tested.
+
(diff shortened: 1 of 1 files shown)


snapshot_manager/snapshot_manager/snapshot_manager.py
```diff
@@ -485,3 +485,256 @@ def handle_labels(
         )
         logging.info(f"Adding label: {labels_to_add}")
         issue.add_to_labels(*labels_to_add)
+
+
+def run_performance_comparison(
+    conf_a: config.Config,
+    conf_b: config.Config,
+    github_repo: str,
+    copr_client: copr.v3.Client,
```
tuliom, line 494: The failing test mention this line. Are we missing an `import` statement?

tuliom (changes requested):
It looks like this PR is missing an import statement that is causing 4 checks to fail.

snapshot_manager/snapshot_manager/snapshot_manager.py
kwk: Sorry to say this but you should ignore failing tests in the stack if it isn't the top-most PR that you're reviewing.

nikic: Shouldn't tests pass at every PR in the stack?

kwk: > Shouldn't tests pass at every PR in the stack?

It depends on what we want. I explained myself in a different comment that I didn't think it was necessary. It helps with git bisect, yes, but it helps less with understanding the "flow" of the commits from the outside to the inside. But since it caused to many questions, I'll never do this again.

PR #1281, opened 2025-04-10
Refactor and add make_snapshot_gating_request

kwk:
* __->__ #1281

There was more and more testing-farm related code in
`testing_farm_util.py` that it made sense to split it up into multiple
modules.

Some functionality that is shared between modules is located in
`tfutil.py`.

The `make_compare_compile_time_request` is probably the most important
change in this PR because that is the one that will kick off the
`compare-compile-time.fmf` plan to run. Tests for this function have
already been added in a previous PR entitled "[tests] Entrypoint fns for
run + collect"

I didn't mention this before but PRs before this one actually made
use of this refactoring. For example the "perf-diagrams" command added
before is adding an extra column `testing_farm_artifacts_url` to its
internal `pandas.DataFrame` that required information from testing-farm.


snapshot_manager/testing_farm/tfutil.py
```diff
@@ -0,0 +1,411 @@
+"""TBD"""
```
tuliom, line 1: Should this be removed?

snapshot_manager/testing_farm/watch_result.py
```diff
@@ -0,0 +1,175 @@
...
+
+    @classmethod
+    def is_watch_result(cls, string: str) -> bool:
+        """_summary_
+
+        Args:
+            string (str): _description_
+
+        Returns:
+            bool: _description_
```
tuliom, line 108: The usage of `_summary_` and `_description_` is coming from the old file.
I wonder if this is a good time to replace them.

tuliom (changes requested):
The implementation LGTM, but I found 2 cosmetic issues.
One of them was already present in the file that got removed.

snapshot_manager/testing_farm/watch_result.py
kwk: Fixed in 449cba921e034d0304dbc13f07f1c7bc4d426fb8

snapshot_manager/testing_farm/tfutil.py
kwk: Fixed in 449cba921e034d0304dbc13f07f1c7bc4d426fb8

PR #1287, opened 2025-04-11
Use unpriviledged user in mass rebuild container

tstellar:
Some packages, like vlc have tests that can't run as root.


tuliom (approved):

tstellar:
I think I found a bug with this change, so I'm not going to merge it yet.

PR #1300, opened 2025-04-16
mass-rebuild-reporter: Rebuild failed packages in COPR

tstellar:
Before we report a failure as a regression, rebuild it in COPR to make sure that it is a true failure.  The benefit of doing this in COPR instead of locally, is that if the build passes then it will become part of the baseline packages for the next rebuild.  Otherwise, if the latest build in COPR is still a failure, then the package won't be rebuilt during the next mass rebuild.

### DIFF
--- a/.github/workflows/mass-rebuild-reporter.yml
+++ b/.github/workflows/mass-rebuild-reporter.yml
@@ -75,8 +75,9 @@ jobs:
           echo "REGRESSIONS=$(cat regressions)" >> "$GITHUB_OUTPUT"
 
   rebuild-failures:
-    if: >- github.repository_owner == 'fedora-llvm-team' &&
-           needs.check-for-rebuild.outputs.rebuild-completed == 'true'
+    if: >-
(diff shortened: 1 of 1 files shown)


.github/workflows/mass-rebuild-reporter.yml
```diff
@@ -75,8 +74,66 @@ jobs:
           python3 scripts/rebuilder.py get-regressions --start-date ${{ steps.last-report.outputs.result }} > regressions
           echo "REGRESSIONS=$(cat regressions)" >> "$GITHUB_OUTPUT"
 
+  rebuild-failures:
+    if: >- github.repository_owner == 'fedora-llvm-team' &&
```
tuliom, line 78: Extra `-` in `>-`?

tuliom (changes requested):
@tstellar It looks like there are 3 issues reported by the pre-commit workflow.

kwk (approved):

tuliom (approved):

PR #1307, opened 2025-04-17
Adapt to new model of requests array

kwk:
Previously it was a dictionary with the chroot as key.

### DIFF
--- a/snapshot_manager/snapshot_manager/snapshot_manager.py
+++ b/snapshot_manager/snapshot_manager/snapshot_manager.py
@@ -280,16 +280,16 @@ class SnapshotManager:
             # But if a package was re-build and failed, the old request ID for that chroot is invalid.
             # To compensate for this scenario that we saw on April 1st 2024 btw., we're gonna
             # delete any request that has a different set of Copr build IDs associated with it.
-            if chroot in requests:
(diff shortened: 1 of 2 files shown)


snapshot_manager/snapshot_manager/snapshot_manager.py
```diff
@@ -280,16 +280,15 @@ def check_todays_builds(self) -> None:
             # But if a package was re-build and failed, the old request ID for that chroot is invalid.
             # To compensate for this scenario that we saw on April 1st 2024 btw., we're gonna
             # delete any request that has a different set of Copr build IDs associated with it.
-            if chroot in requests:
-                recovered_request = requests[chroot]
-                if set(recovered_request.copr_build_ids) != set(current_copr_build_ids):
+            if request in requests:
```
tuliom, line 283: Where is `request` being assigned? I don't see this variable being used anywhere in this function.

tuliom:
I believe that line 307 also need to be modified.

snapshot_manager/snapshot_manager/snapshot_manager.py
kwk: You're right.

tuliom (approved):

PR #1384, opened 2025-05-05
rebuilder: Start using rawhide chroots

tstellar:
rawhide is the only Fedora release that we have s390x builds for, so we need to start using this for doing the rebuilds.

Also factor out the code for building a package, so we can reuse it in future changes to the script.


tuliom (approved):
LGTM, but I think it will require a rebase before merging it.

PR #1477, opened 2025-05-28
Fix errors caused by very long Github comments

tuliom:
Avoid printing comments that are longer than Github's 65536 limit. This limit is rarely passed, but when it is, it causes an exception that abort the execution of the Github Action.
In the rare cases this happens, it's better to have just a link to the build log and a successful execution than aborting everything.


kwk (approved):
It's not tested but it should be fine.

PR #1498, opened 2025-06-11
Submit failed tests to log-detective

kwk:
With `/submit-to-log-detective <CHROOT> [CHROOT ...]` members of the
github `llvm-toolset-engineers` org team can now push pre-annotated
build logs to log-detective.

Here's how the workflow looks in my test repo:

image

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,8 +1,0 @@
-[mypy]
-plugins = pydantic.mypy
-
-[pydantic-mypy]
-init_forbid_extra = True
-init_typed = True
-warn_required_dynamic_aliases = True
-warn_untyped_fields = True


tuliom (approved):
LGTM, but before this is merged, please check if the commit message will be correct.
There is an attached image there that may not work/be available after this is committed.

kwk:
> LGTM, but before this is merged, please check if the commit message will be correct. There is an attached image there that may not work/be available after this is committed.

@tuliom thank you for pointing this out. I only attached the image to the initial comment body after the PR was created. In the final commit message. this image doesn't appear: https://github.com/fedora-llvm-team/llvm-snapshots/commit/ea2931afe7f0fef6e41351bf04f5964d20f27610

PR #1610, opened 2025-08-29
Handle 404 on build log file

nikic:
It's possible for a build log URL to be listed, but the actual lookup to 404, see: https://copr.fedorainfracloud.org/coprs/build/9498339

Treat this the same way as a missing build log, rather than failing the check snapshot task.

### DIFF
--- a/snapshot_manager/snapshot_manager/build_status.py
+++ b/snapshot_manager/snapshot_manager/build_status.py
@@ -8,6 +8,8 @@ import logging
 import pathlib
 import urllib
 
+import requests
+
 import snapshot_manager.util as util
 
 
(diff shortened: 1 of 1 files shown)


snapshot_manager/snapshot_manager/build_status.py
```diff
@@ -226,23 +231,46 @@ def augment_with_error(self) -> "BuildState":
...
+
+        self.err_cause, self.err_ctx, self._err_orig_ctx = get_cause_from_build_log(
+            build_log_file=self._build_log_file
+        )
+
+        return self
+
+    def report_missing_build_log(self) -> "BuildState":
+        # Treat errors with no build logs as unknown and tell user to visit the
+        # build URL manually.
```
kwk, line 259:
```suggestion
        """ Treat errors with no build logs as unknown and tell user to visit the
            build URL manually. """
```

kwk (dismissed):
Thank you for the fix. Just a small documentation adjustment. Otherwise this looks good.

PR #1611, opened 2025-08-29
Use exit_code variable

nikic:
proc.returncode is not set in the timeout case, use the exit_code variable instead.


snapshot_manager/snapshot_manager/util.py
```diff
@@ -128,7 +128,7 @@ def run_cmd(cmd: str, timeout_secs: int | None = 5) -> tuple[int, str, str]:
 
     if exit_code != 0:
         logging.debug(
-            f"exit code: {proc.returncode} for cmd: {cmd}\n\nstdout={stdout}\n\nstderr={stderr}"
+            f"exit code: {exit_code} for cmd: {cmd}\n\nstdout={stdout}\n\nstderr={stderr}"
```
kwk, line 131: The exit code in the timeout case is always -1. Maybe we should change this to the [timeout](https://man7.org/linux/man-pages/man1/timeout.1.html) return code of `124`:

>       124    if COMMAND times out, and --preserve-status is not
>              specified

In other words: I suggest to adjust the `exit_code = -1` above to `exit_code = 124`.

kwk (changes requested):

PR #1623, opened 2025-09-04
Increase the priority of the copr repos when testing

### LINKED ISSUE #395: Evaluate if we should adjust the copr repo priority in testing-farm
In #393 and #394 we've lowered the priority of the `testing-farm-tag-repository`. We might want to adjust the copr repo's priority instead.

tuliom:
Instead of decreasing the priority of the Testing Farm repositories,
increase the priority of our Copr repositories in order to guarantee
that packages from the daily snapshots are preferred.

This also fix an issue caused by Testing Farm repositories changing
names recently.

Fixes #395.

### DIFF
--- a/tests/snapshot-gating.fmf
+++ b/tests/snapshot-gating.fmf
@@ -10,21 +10,11 @@
 
 summary: LLVM Tests for snapshot gating
 prepare:
-  # Lower the priority of the testing-farm-tag-repository so that our copr repo is picked up.
-  # See: https://docs.testing-farm.io/Testing%20Farm/0.1/test-environment.html#_tag_repository
-  - name: Set testing-farm-tag-repository priority
-    how: shell
-    script: |
-      if dnf repolist | grep -q testing-farm-tag-repository; then
-        dnf install -y 'dnf5-command(config-manager)' || dnf install -y 'dnf-command(config-manager)'
-        dnf config-manager --save --setopt="testing-farm-tag-repository.priority=999" || \
-          dnf config-manager setopt "testing-farm-tag-repository.priority=999"
-      fi
-
   - name: Enable copr repo
     how: shell
     script:
       - dnf install -y 'dnf5-command(copr)' || dnf install -y 'dnf-command(copr)'
+      - dnf install -y 'dnf5-command(config-manager)' || dnf install -y 'dnf-command(config-manager)'
       - dnf -y copr enable @fedora-llvm-team/$COPR_PROJECT $COPR_CHROOT
       # dnf5's copr plugin has trouble resolving the runtime dependency repos,
       # and is not replacing the $distname special var. Hence we need to do that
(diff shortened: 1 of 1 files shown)


kwk:
@tuliom do you know if DNF5 behaves the same way? I think you can test this manually in a RHEL8 environment or something alike:

```console
$ podman run -it --rm ubi8
[root@5698718a216e /]# cat /etc/redhat-release 
Red Hat Enterprise Linux release 8.10 (Ootpa)
[root@5698718a216e /]# dnf install -yq 'dnf-command(config-manager)'

Installed:
  dnf-plugins-core-4.0.21-25.el8.noarch                                                                                                      
```

Now show that no priority is currently set and then set one:

```console
[root@5698718a216e /]# grep "priority" /etc/yum.repos.d/ -Hnr
[root@5698718a216e /]# dnf config-manager --save --setopt="ubi-8-baseos-debug-rpms.priority=999"
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use subscription-manager to register.

[root@5698718a216e /]# grep "priority" /etc/yum.repos.d/ -Hnr
/etc/yum.repos.d/ubi.repo:15:priority=999
```

Does this help? AFAIU you don't need the additional trailing repo. It is enough to have it in the string, no?

nikic (commented):
I think the problem here is as the error message indicates: The repository does not exist.

This is probably related to https://gitlab.com/testing-farm/infrastructure/-/merge_requests/1061/diffs. I think the name would now be testing-farm-tag-repository-mainline?

tests/snapshot-gating.fmf
```diff
@@ -37,6 +27,8 @@ prepare:
       - dnf -y install --best llvm-libs
       # Use a newer llvm-test-suite version from copr.
       - dnf copr enable -y @fedora-llvm-team/llvm-test-suite $COPR_CHROOT
+      # Increase the priority of the copr repositories in order to prefer packages from them.
+      - dnf config-manager --save --setopt="copr:copr.fedorainfracloud.org:group_fedora-llvm-team:*.priority=50" || dnf config-manager setopt "copr:copr.fedorainfracloud.org:group_fedora-llvm-team:*.priority=50"
```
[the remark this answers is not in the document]
tuliom: Nice catch! I've just fixed this.

tests/snapshot-gating.fmf
```diff
@@ -37,6 +27,10 @@ prepare:
       - dnf -y install --best llvm-libs
       # Use a newer llvm-test-suite version from copr.
       - dnf copr enable -y @fedora-llvm-team/llvm-test-suite $COPR_CHROOT
+      # Increase the priority of the copr repositories in order to prefer packages from them.
+      # Testing Farm uses priority=9 by default.
+      # Source: https://docs.testing-farm.io/Testing%20Farm/0.1/test-environment.html#_tag_repository
+      - dnf config-manager --save --setopt="copr:copr.fedorainfracloud.org:group_fedora-llvm-team:*.priority=8" || dnf config-manager setopt "copr:copr.fedorainfracloud.org:group_fedora-llvm-team:*.priority=8"
```
nikic, line 33: Did you test locally whether this common works after enabling the copr repo?

nikic (approved):
LGTM. I think it's okay to just land it, we'll see if it works soon enough :)

tests/snapshot-gating.fmf
tuliom: I did run manually the DNF5 command in a container configured by hand.
It worked.

I'm afraid the actual environment is far more diverse, though.

PR #1747, opened 2025-10-30
Drop with-flang strategy

kwk:
We can drop the `with-flang` build strategy now.

I intentionally left it for today after [landing `flang`](https://src.fedoraproject.org/rpms/llvm/c/ee920d4db95425b764c576fa7ad97b3ae86038fb?branch=rawhide) in the `rpms/llvm` repository so that I can see if any error happening today failed in my branch (with-flang strategy) or in big-merge as well.

The packaging error appeared in both build strategies (https://github.com/fedora-llvm-team/llvm-snapshots/issues/1745#issuecomment-3466178323).

### DIFF
--- a/.github/workflows/fedora-copr-build.yml
+++ b/.github/workflows/fedora-copr-build.yml
@@ -16,7 +16,6 @@ on:
         options:
         - all
         - big-merge
-        - with-flang
 
 jobs:
   generate-matrix:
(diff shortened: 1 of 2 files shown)


nikic:
(Though is it possible to make that just a text input without a fixed set of options? Shouldn't repeat this in two places...)

PR #1771, opened 2025-11-11
Upload paths matching: llvm*-YYYYMMDD.txt

kwk:
Before the `llvm-test-suite-version-<YYYYMMDD>.txt` file wasn't uploaded.

Fixup for #1766


jchecahi (approved):
The glob makes sense to me

PR #1827, opened 2025-12-09
Stop forking the llvm-test-suite repository

tuliom:
Treat the llvm-test-suite repository differently by building its packages directly in the final repository, instead of creating a new repository that is forked into the final repository.

Fixes #1815.

PR #1830, opened 2025-12-10
Fix promotion of llvm-test-suite-builds

### LINKED ISSUE #1815: llvm-test-suite builds stopped working
The last build is from Nov. 05th.
There have been 6 commits to the repository after that.

kwk:
This implements the fix I've addressed [here](https://github.com/fedora-llvm-team/llvm-snapshots/pull/1827#issuecomment-3636293612):

> I think the logic I've used to implemented is error-prone. IIRC it said, take a look at yesterday's git rev and LLVM rev. If they are the same, then don't build llvm-test-suite. This makes sense at first but if the build from yesterday was new and needs a promotion we need to look at the day before yesterday. The logic to promote a llvm-test-suite build should be this:
> 
> If `git_rev_yesterday != git_rev_day_before_yesterday || llvm_version_yesterday != llvm_version_day_before_yesterday`, then promote yesterday's build of llvm-tests-suite. Make sense?
> 
> It probably makes sense to have a dedicated `promote_build` variable instead of this ugly `yesterdays_project_exists` condition that implies to promote a build.
> 

Fixes #1815

### DIFF
--- a/.github/workflows/fedora-copr-build.yml
+++ b/.github/workflows/fedora-copr-build.yml
@@ -67,14 +67,17 @@ jobs:
         run: |
           today=$(date +%Y%m%d)
           yesterday=$(date -d "${today} -1 day" +%Y%m%d)
+          day_before_yesterday=$(date -d "${yesterday} -1 day" +%Y%m%d)
 
           username=@fedora-llvm-team
           {
             echo "username=$username"
-            echo "yesterday=$yesterday"
             echo "today=$today"
+            echo "yesterday=$yesterday"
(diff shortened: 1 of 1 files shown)


tuliom:
I have no objections to this, but it would be great to confirm if we do have consensus on this before merging.

kwk:
As discussed in meeting with @tuliom and @nikic we're going with https://github.com/fedora-llvm-team/llvm-snapshots/pull/1827 for now. Closing this. Maybe we have to resurrect this PR later.

PR #1831, opened 2025-12-10
scripts/rebuilder.py: Add support for testing clang-built packages

tstellar:
This adds a new command called 'test', which will test all the clang-built packages in CentOS Stream 9, CentOS Stream 10, and Fedora Rawhide.  It will use the latest successful snapshot build for each arch/OS combination to do the builds.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ copr==2.5
     #   copr-cli
 copr-cli==2.4
     # via -r requirements.txt.in
-coverage==7.13.1
+coverage==7.13.2
     # via -r requirements.txt.in
 cryptography==44.0.1
     # via pyjwt
--- a/requirements.txt.in
+++ b/requirements.txt.in
@@ -13,6 +13,6 @@ munch==4.0.0
 copr==2.5
 requests==2.32.5
 fnc
-coverage==7.13.1
+coverage==7.13.2
 koji
 pytest==9.0.2


tuliom (commented):
I have a suggestion for improvement, but the code LGTM.

scripts/rebuilder.py
```diff
@@ -519,6 +634,134 @@ def main() -> None:
...
+                print(f"Skip building {pkg} on rawhide, because it is FTBFS")
+                continue
+            print(f"Building {pkg}")
+            build_pkg(
+                project_owner,
+                project_name,
+                copr_client,
+                pkg,
+                default_commitish="rawhide",
+                build_tag="f44",
```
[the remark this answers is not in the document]
tstellar: Good idea, I've implemented this.

tuliom (dismissed):

scripts/rebuilder.py
```diff
@@ -14,6 +15,12 @@
 from munch import Munch
 
 
+def get_rawhide_tag() -> Any:
+    koji_session = koji.ClientSession("https://koji.fedoraproject.org/kojihub")
+    target = koji_session.getBuildTarget("rawhide")
+    return target["build_tag_name"].split("-")[0]
```
kwk:
```suggestion
def get_rawhide_tag() -> str:
    """ Returns the current tag for rawhide, i.e. "f44". """
    koji_session = koji.ClientSession("https://koji.fedoraproject.org/kojihub")
    target = koji_session.getBuildTarget("rawhide")
    return target["build_tag_name"].split("-")[0]
```

I suggest to make the return type annotation `str` because of this:

```python
>>> import koji
>>> koji_session = koji.ClientSession("https://koji.fedoraproject.org/kojihub")
... target = koji_session.getBuildTarget("rawhide")
... type(target["build_tag_name"].split("-")[0])
... 
<class 'str'>
```

scripts/rebuilder.py
```diff
@@ -27,6 +34,88 @@ def is_tier0_package(pkg: str) -> bool:
...
+        )
+    repos = base.repos.get_matching("*")
+    repos.disable()
+    repos = base.repos.get_matching("*-source*")
+    repos.enable()
+
+    base.fill_sack()
+    q = base.sack.query(flags=hawkey.IGNORE_MODULAR_EXCLUDES)
+    q = q.available()
+    q = q.filter(requires=["clang"])
```
kwk: Does are `require` key really identifies the dependency at build-time as the documentation of this function suggests? `# Packages in CentOS Stream that are built by clang`

Shouldn't this be search for `biuld_requires`?

tstellar: There is no option for build_requires as far as I  know.  This works because it is querying a sources repo.

scripts/rebuilder.py
tstellar: I had this as returning `str` originally, but the pre-commit job flagged it as an error.  I can try again.

scripts/rebuilder.py
```diff
@@ -14,6 +15,13 @@
 from munch import Munch
 
 
+def get_rawhide_tag() -> str:
+    """Returns the current tag for rawhide, i.e. "f44"."""
+    koji_session = koji.ClientSession("https://koji.fedoraproject.org/kojihub")
+    target = koji_session.getBuildTarget("rawhide")
```
tuliom: I wonder if this should be rewritten as:
```
target: str = koji_session.getBuildTarget("rawhide")
```

tuliom (dismissed):

PR #1899, opened 2026-01-28
Correctly handle riscv64 chroots

kwk:
* Add missing arch to allowed_archs_as_regex_str
  The `chroot_pattern` in `config.py` couldn't be filtered to remove
  `riscv64` chroots before this patch because
  it thought that is was an unknown architecture.

* While riscv64 indeed IS a known architecutre now,
  we still don't want to build for it.
  When sanitizing the chroots to be used in the config
  we remove all riscv64 chroots.

### DIFF
--- a/snapshot_manager/snapshot_manager/util.py
+++ b/snapshot_manager/snapshot_manager/util.py
@@ -261,9 +261,9 @@ def allowed_archs() -> list[str]:
     Example:
 
     >>> sorted(allowed_archs())
-    ['aarch64', 'i386', 'ppc64le', 's390x', 'x86_64']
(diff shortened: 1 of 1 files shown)


snapshot_manager/snapshot_manager/util.py
```diff
@@ -272,7 +272,7 @@ def allowed_archs_as_regex_str() -> str:
     Example:
 
     >>> allowed_archs_as_regex_str()
-    '(aarch64|i386|ppc64le|s390x|x86_64)'
+    '(aarch64|i386|ppc64le|s390x|x86_64|riscv64|x86_64_v2)'
```
nikic: Does allowed_archs need to be updated as well?

Also, what effect is this going to have? Are we doing to start building for these chroots as well?

PR #1955, opened 2026-03-09
Update tft-cli

nikic:
The currently used version no longer worked due to removal of pkg_resources.

We pinned the version in https://github.com/fedora-llvm-team/llvm-snapshots/pull/633. I don't know if that issue still exists, but the current version just doesn't work at all.


kwk:
@nikic which version doesn't work at all? The pinned `0.0.16` or a newer one than `0.0.32`?

PR #2020, opened 2026-04-21
Update snapshot maintainer

tbaederr:


tuliom:
@tbaederr Has there been a change in the schedule?

tbaederr:
No, what's the schedule? We have a new sprint now, so a new maintainer. Not sure why you're the maintainer right now since you're doing Fedora packaging.

nikic:
It's @tstellar per the schedule, but he was still out on Tuesday and we didn't know when he was coming back.

tuliom:
This is my suggestion: close this PR and wait for the next scheduled slot for you.
I can continue covering for @tstellar while he's catching up.

Unless you'd like to change the schedule. If that's the case, we can review it.

PR #2076, opened 2026-05-27
Update the actions/setup-python version

kwk:
Update the actions/setup-python version.

### DIFF
--- a/.github/actions/prepare-python/action.yml
+++ b/.github/actions/prepare-python/action.yml
@@ -22,16 +22,6 @@ runs:
         fi
         echo "is-fedora=$fedora" >> $GITHUB_OUTPUT
 
-    - name: Install python setup prerequisites
-      shell: bash
-      run: |
-        if [ "${{ steps.os-check.outputs.is-fedora }}" = "true" ]; then
-          sudo dnf install -y lsb_release
-        else
-          sudo apt-get update
-          sudo apt-get install -y lsb-release
-        fi
-
     - name: "Setup python"
       uses: actions/setup-python@v6.2.0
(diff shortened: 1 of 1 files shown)


.github/actions/prepare-python/action.yml
```diff
@@ -53,7 +63,7 @@ runs:
     - name: "Install python dependencies for project"
       shell: bash
       run: |
-        pip install -r ${{ inputs.checkout-path }}/requirements.txt
+        python -m pip install -r ${{ inputs.checkout-path }}/requirements.txt
```
tuliom: Earlier in this file we have:
```
pip install pytest coverage
```

Here we're replacing `pip install ...` with `python -m pip install ...`.

1. What is the difference between `pip install ...` and `python -m pip install ...`? Is one of them ensuring that a specific python version is used?
2. Depending on the answer to the previous question, do we need to modify how `pytest` and `coverage` are installed too? And why?
3. Could we add a source comment describing why we're doing this? e.g.:

```
# Use 'python -m pip' because our tests require a specific python version which was setup via actions/setup-python.
```

kwk: I've tried some things and kept everything until it worked. But with the new change less things are required. Only the cache line needed to drop and I've updated the action version. Sorry for the big change before.

kwk: Okay, even the cache can come back now. Great. There was an error with dependencies before for every new PR and now this works and should unblock #2012 .

tuliom (approved):
LGTM.

Can dependabot also update these actions?

kwk:
> LGTM.
> 
> Can dependabot also update these actions?

I don't know TBH.

PR #2083, opened 2026-05-27
Configuring multi-ecosystem updates for Dependabot

kwk:
This should reduce the number of Dependabot pull requests we receive by grouping updates across multiple ecosystems into a single, consolidated pull request.

See [Configuring multi-ecosystem updates for Dependabot](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/configuring-multi-ecosystem-updates)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,8 +27,6 @@ updates:
       - "pip"
       - "python"
     target-branch: "main"
-    reviewers:
-      - "kwk"
     patterns: ["*"]
     multi-ecosystem-group: "infrastructure"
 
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,11 +4,18 @@
 # https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
 version: 2
+
+multi-ecosystem-groups:
+  infrastructure:
+    schedule:
+      interval: "daily"
```
kwk, line 11: Before you ask, I want to have all dependabot updates to be done daily, yes :)

tuliom (dismissed):
I suspect the `reviewers` part should be moved or removed.
But LGTM as-is.

kwk:
> I suspect the `reviewers` part should be moved or removed. But LGTM as-is.

You're correct. I don't know when I added this but it apparently never worked.

tuliom (approved):

PR #2102, opened 2026-05-29
Added scripts/package-docs-for-rhel.sh

### LINKED ISSUE #2090: Download all RPMs for a particular NVR for a given architecutre (e.g. x86_64)
Make sure to include noarch RPMs as well.

### LINKED ISSUE #2091: Download SRPM for a given NVR to get access to the spec file used in that version.

### LINKED ISSUE #2092: Extract the SRPM (NOT only the llvm.spec) to some directory.

### LINKED ISSUE #2093: Parse the spec file for /usr/share/man files
`rpmspec --undefine=fedora --define='rhel 9' -P llvm.spec | grep -P '^/usr/share/man'`

kwk:
Fixes #2090 - Download all RPMs for a particular NVR for a given architecutre (e.g. x86_64)
Fixes #2091 - Download SRPM for a given NVR to get access to the spec file used in that version.
Fixes #2092 - Extract the SRPM (NOT only the llvm.spec) to some directory.
Fixes #2093 - Parse the spec file for /usr/share/man files

### DIFF
--- a/scripts/package-docs-for-rhel.sh
+++ b/scripts/package-docs-for-rhel.sh
@@ -99,6 +99,7 @@ function main {
     echo "INFO: Target RHEL version: ${rhel_version}" 1>&2
 
     download_rpms "${nvr}" "${arch}" "${base_dir}"
+    download_rpms "${nvr}" "noarch" "${base_dir}"
     download_srpm "${nvr}" "${base_dir}"
     extract_srpm "${base_dir}/${nvr}.src.rpm" "${base_dir}/srpm"
     parse_spec_file_for_doc_files "${base_dir}/srpm/llvm.spec" "${rhel_version}" > "${doc_files}"
--- a/scripts/package-docs-for-rhel.sh
+++ b/scripts/package-docs-for-rhel.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+
+set -e
+
+# Prints whatever tag is currently assigned to rawhide in koji (e.g. f45).
+function get_rawhide_tag {
+    local rawhide_tag
+
+    rawhide_tag=$(koji list-targets --name=rawhide --quiet | grep -oP "f[0-9]+" | head -n1)
+    echo "INFO: rawhide tag: ${rawhide_tag}" 1>&2
+    echo "$rawhide_tag"
+}
+
+# Prints the latest NVR (e.g. llvm-22.1.6-1.fc45) of the llvm package for the
+# given rawhide tag (e.g. f45). If no rawhide tag is provided, it will be
+# determined automatically.
+function get_nvr {
+    local rawhide_tag=${1:-$(get_rawhide_tag)}
+    local nvr
+
+    nvr=$(koji latest-build --quiet "${rawhide_tag}" llvm | cut -d ' ' -f1)
+    echo "INFO: NVR: ${nvr}" 1>&2
+    echo "$nvr"
+}
+
+# Downloads the RPMs for a given NVR and architecture into a target directory.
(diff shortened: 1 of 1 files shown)


scripts/package-docs-for-rhel.sh
```diff
@@ -0,0 +1,107 @@
...
+# Downloads the RPMs for a given NVR and architecture into a target directory.
+function download_rpms {
+    local nvr=${1:-$(get_nvr)}
+    local arch=${2:-x86_64}
+    local target_dir=${3:-./$PWD/${arch}}
+
+    echo "INFO: Downloading RPMS for ${nvr} and arch ${arch} to ${target_dir}" 1>&2
+    mkdir -p "${target_dir}"
+    pushd "${target_dir}"
+    koji download-build --arch="${arch}" "${nvr}"
```
tuliom, line 42: We need to download `noarch` packages too

scripts/package-docs-for-rhel.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
+
+set -e
+
+# Prints whatever tag is currently assigned to rawhide in koji (e.g. f45).
+function get_rawhide_tag {
+    local rawhide_tag
+
+    rawhide_tag=$(koji list-targets --name=rawhide --quiet | grep -oP "f[0-9]+" | head -n1)
```
tuliom: It took me a while to understand why `koji list-tags` was not used here.
I believe it's worth mentioning it.

Perhaps something like this:

```
# list-targets provides us a mapping between rawhide -> targets.
# We assume that target names are prefixed with the name of the tags.
```

scripts/package-docs-for-rhel.sh
```diff
@@ -0,0 +1,107 @@
...
+    | grep -P '^/usr/share/man'
+    popd
+}
+
+function main {
+    local rawhide_tag
+    local nvr
+    local arch=x86_64
+    local base_dir="${PWD}/package-rhel-docs"
+    local final_dir="${base_dir}/final"
```
tuliom: Where is this directory used?

tuliom (changes requested):

scripts/package-docs-for-rhel.sh
kwk: Done in 97343f805fedceed762b1cd3474d92ecb74b2f06.

scripts/package-docs-for-rhel.sh
kwk: This code was inspired by code that already exists in the repo [here](https://github.com/fedora-llvm-team/llvm-snapshots/blob/5043a9b0d05bd828e26c0cfdf78b7f97f0f34542/scripts/rebuilder.py#L28-L41): 

```python
def get_rawhide_tag() -> str:
    """Returns the current tag for rawhide, i.e. "f44".


    Example:

    >>> tag=get_rawhide_tag()
    >>> bool(re.match("f[0-9]+", tag))
    True
    """
    koji_session = koji.ClientSession("https://koji.fedoraproject.org/kojihub")
    target = koji_session.getBuildTarget("rawhide")
    build_tag_name: str = target["build_tag_name"]
    return build_tag_name.split("-")[0]
```

I find it pretty straight-forward. But I get your thought. I've added a comment in b4c92570.

scripts/package-docs-for-rhel.sh
kwk: Not used and removed in 06a97f5.

tuliom (approved):

PR #2126, opened 2026-06-17
Build and upload man pages alongside other snapshot version information

kwk:
Other changes included are:

* Remove unused and deprecated file: `scripts/package-docs-for-rhel.sh`

* I also ran `precommit autoupdate` because without that I ran into this error which to me looks like an error not on my side.

```    pyupgrade................................................................Failed
    - hook id: pyupgrade
    - exit code: 1

    Traceback (most recent call last):
      File "/home/kkleine/.cache/pre-commit/repo68ioyw5f/py_env-python3.14/bin/pyupgrade", line 8, in <module> sys.exit(main()) ~~~~^^ File "/home/kkleine/.cache/pre-commit/repo68ioyw5f/py_env-python3.14/lib/python3.14/site-packages/pyupgrade/_main.py", line 389, in main ret |= _fix_file(filename, args) ~~~~~~~~~^^^^^^^^^^^^^^^^ File "/home/kkleine/.cache/pre-commit/repo68ioyw5f/py_env-python3.14/lib/python3.14/site-packages/pyupgrade/_main.py", line 327, in _fix_file contents_text = _fix_tokens(contents_text) File "/home/kkleine/.cache/pre-commit/repo68ioyw5f/py_env-python3.14/lib/python3.14/site-packages/pyupgrade/_main.py", line 297, in _fix_tokens tokenize.cookie_re.match(token.src) ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^ TypeError: cannot use a bytes pattern on a string-like object
```

* Fixes a deprecation warning while at it:

```
    DeprecationWarning: Argument login_or_token is deprecated, please use
    auth=github.Auth.Token(...) instead
```


nikic:
> Don't waste time searching for a good commit. Just run one check and then take main.

I don't understand why this change is made? (If we want to make it, we should do it separately, and remove the good commit finding entirely. Doesn't make much sense to find a good commit among a list of one...)

kwk:
> > Don't waste time searching for a good commit. Just run one check and then take main.
> 
> I don't understand why this change is made? (If we want to make it, we should do it separately, and remove the good commit finding entirely. Doesn't make much sense to find a good commit among a list of one...)

When running the workflow and sitting next to the output I noticed that we still search for a special builder to succeed that no longer exists (?) in upstream llvm. We eventually always fall back to HEAD. I agree that this has absolutely nothing to do with this PR and could be done in another one. Let me create that.

nikic (approved):
LGTM

PR #2127, opened 2026-06-17
Change the check we look for in a good commit from clang-x86_64-debian-fast to release-noassertions-warnings

kwk:
The former, `clang-x86_64-debian-fast`, no longer exists upstream.


nikic (approved):
LGTM

PR #2148, opened 2026-07-03
Adjust check-snapshots scheduling

nikic:
This used a weird `*/60` syntax, which I *guess* is probably equivalent to just `0`? Scheduling at `0` is not recommended because it has particularly high load. Pick a random other minute value instead.


kwk (approved):
Looks good. I usually very things with `https://crontab.guru/#17_*_*_*_*` . That's suggested by the github docs as well.

tuliom (approved):

PR #2149, opened 2026-07-03
Update Fedora version in test

nikic:
F42 is EOL, so the repo metadata is no longer available at that location.

### DIFF
--- a/scripts/rebuilder.py
+++ b/scripts/rebuilder.py
@@ -142,7 +142,7 @@ def get_tier2_pkgs(version: str = "rawhide") -> set[str]:
     """Returns all packages that BuildRequires clang for the given Fedora version
 
     Args:
-        version (str): A Fedora version sring e.g. rawhide, 43, 42, etc.
+        version (str): A Fedora version string e.g. rawhide, 44, 43, etc.
     Returns:
         set[str]: A set of package names.
     Exmaple:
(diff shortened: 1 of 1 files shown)


scripts/rebuilder.py
```diff
@@ -142,15 +142,15 @@ def get_tier2_pkgs(version: str = "rawhide") -> set[str]:
     """Returns all packages that BuildRequires clang for the given Fedora version
 
     Args:
-        version (str): A Fedora version sring e.g. rawhide, 43, 42, etc.
+        version (str): A Fedora version string e.g. rawhide, 44, 43, etc.
```
kwk: I don't think it is needed to update the documentation.

scripts/rebuilder.py
```diff
@@ -142,15 +142,15 @@ def get_tier2_pkgs(version: str = "rawhide") -> set[str]:
...
     Returns:
         set[str]: A set of package names.
     Exmaple:
 
     >>> pkgs=get_tier2_pkgs()
     >>> len(pkgs) > 0
     True
-    >>> pkgs=get_tier2_pkgs("42")
+    >>> pkgs=get_tier2_pkgs("44")
```
kwk: If we want to avoid fixing this in the future we might as well get whatever rawhide's version currently is and use the version before. The `get_rawhide_tag` returns the version with an `f`-prefix. That's why we have to strip of the first character (`[1:]`). I hope this makes sense?


```suggestion
    >>> pkgs=get_tier2_pkgs(int(get_rawhide_tag()[1:])-1)
```

nikic: This doesn't really seem suitable for an example?

kwk: I think it is.

nikic: Okay, done.

kwk (approved):

PR #2166, opened 2026-07-17
Stop looking for tests at tests/flang

tuliom:
We ended up not implementing tests in this repository. Everything we have for flang is available at tests/llvm.


nikic (approved):

kwk:
Good catch.

PR #2232, opened 2026-09-15
Update maintainer

tuliom:


nikic:
Given that the sprint has been extended, maybe I should keep maintaining snapshots for another week as well?

tuliom:
> Given that the sprint has been extended, maybe I should keep maintaining snapshots for another week as well?

@nikic That's a good point. We didn't discuss this. What if I wait until Friday to merge this?

nikic:
Sounds good to me

nikic (approved):

